PR #5, opened 2020-05-16
Last-minute fixes to prepare for v3.1.4

dscho:
Turns out that my previous PR (#4) did not result in anything that compiles.

So here are two patches that fix the compilation (one because the `newpath` variable went away, the other because binutils were upgraded to a version that no longer supports some of the `bfd_*()` functions).

While at it, I also slipped in two more patches that I think are uncontroversial: one to make `strace --quiet` really quiet, and the other to fix the situation where an older `msys-2.0.dll` is overwritten with a freshly-built `msys0.dll` without also copying the new version of `cygwin-console-helper.exe` (which caused me quite some pain to figure out: MinTTY would simply hang and I needed several days to debug this).


dscho:
I finally managed to build a full `msys2-runtime` package with these patches, see https://github.com/msys2/MSYS2-packages/pull/1953.

PR #7, opened 2020-06-05
popen: call /usr/bin/sh instead of /bin/sh

lazka:
We mount /usr/bin to /bin, but in a chroot this is broken and we
have no /bin, so try to use the real path.

chroot is used by pacman to run install scripts when called with --root
and this broke programs in install scripts calling popen()
(install-info from texinfo for example)

There are more paths hardcoded to /bin in cygwin which might also be broken
in this scenario, so this maybe should be extended to all of them.


lazka:
An alternative would be if we require users to call `mount -o bind /tmp/pacman-root/usr/bin /tmp/pacman-root/bin` first

lazka:
Thanks!

For future reference, I also tried to do a bind mount like `mount -o bind /tmp/root/usr/bin /tmp/root/bin` before using `pacman --root`, but this somehow didn't help in this case.

PR #8, opened 2020-06-25
Cygwin: rewrite cmdline parser

Artoria2e5:
*This is a version of a cygwin patch ported to this repo. See https://cygwin.com/pipermail/cygwin-patches/2020q2/010304.html.*

This commit rewrites the cmdline parser to achieve the following:
* MSVCRT compatibility. Except for the single-quote handling (an
  extension for compatibility with old Cygwin), the parser now
  interprets option boundaries exactly like MSVCR since 2008. This fixes
  the issue where our escaping does not work with our own parsing.
* Clarity. Since globify() is no longer responsible for handling the
  opening and closing of quotes, the code is much simpler.
* Sanity. The GLOB_NOCHECK flag is removed, so a failed glob correctly
  returns the literal value. Without the change, anything path-like
  would be garbled by globify's escaping.

Some clarifications are made in the documentation for when globs are not
expanded.  A minor change was made to insert_file to remove the memory
leak with multiple files.

The change fixes two complaints of mine:
* That cygwin is incompatible with its own escape.[1]
* That there is no way to echo `C:\"` from win32.[2]
  [1]: https://cygwin.com/pipermail/cygwin/2020-June/245162.html
  [2]: https://cygwin.com/pipermail/cygwin/2019-October/242790.html

(It's never the point to spawn cygwin32 from cygwin64. Consistency
matters: with yourself always, and with the outside world when you are
supposed to.)

### DIFF
--- a/winsup/cygwin/dcrt0.cc
+++ b/winsup/cygwin/dcrt0.cc
@@ -179,11 +179,14 @@ next_arg (char *&cmd, char *&arg, size_t* quotepos, size_t &quotesize)
 	}
 
       // For anything else, sort out backslashes first.
-      memset (out, '\\', inquote ? nbs / 2 : nbs);
-      out += inquote ? nbs / 2 : nbs;
-
+      // All backslashes are literal, except these before a quote.
       // Single-quote is our addition.  Would love to remove it.
-      if (nbs % 2 == 0 && (inquote ? *cmd == quote : isquote (*cmd)))
+      bool atquote = inquote ? *cmd == quote : isquote (*cmd);
+      size_t n = atquote ? nbs / 2 : nbs;
+      memset (out, '\\', n);
+      out += n;
+
+      if (nbs % 2 == 0 && atquote)
 	{
 	  /* The infamous "" special case: emit literal '"', no change.
 	   *
@@ -328,7 +331,7 @@ build_argv (char *cmd, char **&argv, int &argc, int doglob)
     {
       /* Possibly look for @file construction assuming that this isn't
 	 the very first argument and the @ wasn't quoted */
-      if (argc && quotepos[0] > 1 && *word == '@')
+      if (argc && quotepos[0] > 0 && *word == '@')
 	{
 	  if (++nesting > MAX_AT_FILE_LEVEL)
 	    api_fatal ("Too many levels of nesting for %s", word);
--- a/winsup/cygwin/dcrt0.cc
+++ b/winsup/cygwin/dcrt0.cc
@@ -84,7 +84,7 @@ do_global_ctors (void (**in_pfunc)(), int force)
  * @foo and not the contents of foo.
  */
 static bool __stdcall
-insert_file (char *name, char *&cmd)
+insert_file (const char *name, char *&cmd, int free_old)
 {
   HANDLE f;
   DWORD size;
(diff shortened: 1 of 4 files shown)


winsup/cygwin/dcrt0.cc
```diff
@@ -298,60 +309,31 @@ globify (char *word, char **&argv, int &argc, int &argvlen)
...
-	  if (issep (*cmd))	// End of argument if space
-	    break;
-	}
-      if (*cmd)
-	*cmd++ = '\0';		// Terminate `word'
-
       /* Possibly look for @file construction assuming that this isn't
 	 the very first argument and the @ wasn't quoted */
-      if (argc && sawquote != word && *word == '@')
+      if (argc && quotepos[0] > 1 && *word == '@')
```
Artoria2e5, line 331:
```suggestion
      if (argc && quotepos[0] > 0 && *word == '@')
```

lazka:
* Any reason why we shouldn't wait for this to be included in cygwin upstream?
* Would this help with https://github.com/msys2/MINGW-packages/issues/3548#issuecomment-504441518 ?

Artoria2e5:
Yeah, I am planning to wait for upstream on that. The main reason I am opening a PR is because it helps me read patches faster on a side-by-side diff and binge-pushing on a PR is less of a nuisance compared to binge-pushing on a mailing list. (I am unfortunately the kind of person who yells oh-no after every single email.)

And yes, it would help. Python assumes MSVCRT escapes, so we should aim to be compatible.

winsup/cygwin/dcrt0.cc
```diff
@@ -151,44 +153,71 @@ isquote (char c)
...
+ * Parse a word in-place, consuming characters and marking where quoting state is changed. */
+static /*__inline*/ bool
+next_arg (char *&cmd, char *&arg, size_t* quotepos, size_t &quotesize)
 {
-  char *p;
-  char quote = *cmd;
+  bool inquote = false;
+  size_t nbs = 0;
+  char quote = '\0';
+  quotepos[0] = SIZE_MAX;
```
Artoria2e5, line 164: This should instead be a sentinel always placed at the end. Thanks to insomnia thoughts lmao.

lazka:
ok, thanks for looking into this :)

winsup/cygwin/dcrt0.cc
```diff
@@ -140,6 +140,8 @@ insert_file (char *name, char *&cmd)
 
   tmp[size++] = ' ';
   strcpy (tmp + size, cmd);
+  if (free_old)
```
Artoria2e5, line 143: Can't just do that as cmd does not mean start of the allocated block. Also need to migrate the argv refs if we want to eliminate the strdup (we should). Need some realloc and memmove.

lazka:
I know it's not your fault etc but that there are no tests for this is kinda scary..

Artoria2e5:
Yeah, I was just thinking about that like 10 minutes ago. I am thinking of making a public version of `build_argv` (with `extern "C"`) so that the test suite can access it. As simple as:

```C
char ** cygwin_build_argv(const char* cmdline, int *argc, int doglob) {
    char *cmd = strdup(cmdline);
    int ac;
    char **av;
    build_argv(cmd, av, ac, doglob);
    *argc = ac;
    return av;
}
```

Artoria2e5:
Got it compiling from Cygwin's version of the tree, but the new dll segfaults in crt0. Gotta find out what I fucked up this time…

lazka:
Looks like you got some positive feedback from upstream. Nice!

lazka:
Can we close this?

PR #9, opened 2020-08-09
recursiveCopy: Always close the used handle with FindClose()

lazka:
See https://github.com/msys2/MSYS2-packages/issues/2060


lazka:
reviewed here: https://github.com/msys2/MSYS2-packages/pull/2083#issuecomment-671107777

PR #10, opened 2020-08-11
Fix MSYS=wincmdln, and enable it by default

dscho:
This is apparently broken since 2015...

### DIFF
--- a/winsup/cygwin/globals.cc
+++ b/winsup/cygwin/globals.cc
@@ -68,7 +68,7 @@ bool allow_glob = true;
 bool ignore_case_with_glob;
 bool pipe_byte;
 bool reset_com;
-bool wincmdln;
+bool wincmdln = true;
 winsym_t allow_winsymlinks = WSYM_sysfile;
 bool disable_pcon = true;
 
--- a/winsup/cygwin/spawn.cc
+++ b/winsup/cygwin/spawn.cc
@@ -412,13 +412,13 @@ child_info_spawn::worker (const char *prog_arg, const char *const *argv,
 	          newargv.replace (i, tmpbuf);
 	          free (tmpbuf);
 	        }
-	      if ((wincmdln || !real_path.iscygexec ())
-	            && !cmd.fromargv (newargv, real_path.get_win32 (),
-			        real_path.iscygexec ()))
-	        {
-	          res = -1;
-	          __leave;
-	        }
+	    }
+	  if ((wincmdln || !real_path.iscygexec ())
+	        && !cmd.fromargv (newargv, real_path.get_win32 (),
+		    real_path.iscygexec ()))
+	    {
+	      res = -1;
+	      __leave;
 	    }
 
 
--- a/winsup/doc/cygwinenv.xml
+++ b/winsup/doc/cygwinenv.xml
@@ -68,7 +68,7 @@ time and when handles are inherited.  Defaults to set.</para>
 <listitem>
 <para><envar>(no)wincmdln</envar> - if set, the windows complete command
 line (truncated to ~32K) will be passed on any processes that it creates
-in addition to the normal UNIX argv list.  Defaults to not set.</para>
+in addition to the normal UNIX argv list.  Defaults to set.</para>
 </listitem>
 
 <listitem>


dscho:
I marked this up as a `fixup!` so that the next rebase (provided that `git rebase -i` is used, with auto-squash in effect) squashes these changes into the appropriate patch.

FWIW I wonder whether we should also make `wincmdln` the default for MSYS2?

lazka:
If I understand this correctly this is for non-cygwin binaries in which case the path would have been hit anyway because !real_path.iscygexec(). Am I missing something?

(unrelated, as you might have noticed I've changed the default branch for this repo and added CI. I hope that's ok)

elieux:
Based on my debugging experience, missing command line arguments are an issue in cygwin processes, not Win32 ones.

elieux:
Assuming there's no disadvantage to this feature, I would like this turned on by default (or hard-coded even).

dscho:
> If I understand this correctly this is for non-cygwin binaries in which case the path would have been hit anyway because !real_path.iscygexec(). Am I missing something?

No, this is for MSYS2/Cygwin binaries. If you call, say, `vim.exe abc` from a regular MSYS2 Bash, you won't see the command-line arguments in the output of `wmic process where "CommandLine like '%vim%'" get CommandLine` (see https://github.com/git-for-windows/git/issues/2756 for more details).

> Assuming there's no disadvantage to this feature, I would like this turned on by default (or hard-coded even).

There is only _one_ disadvantage that _I_ can see: as the shown command-line is not the _actual_ command-line as used by MSYS2/Cygwin, there might be discrepancies. The most obvious one comes when the command-line is longer than 32k: this is allowed in MSYS2 when calling another MSYS2 process, but it would be shown truncated by `wmic`.

However, I think you're right: this should be the default nevertheless because the surprise of not seeing any command-line arguments is bigger than the surprise of seeing a potentially truncated one.

Having said that, I do want to give users an "escape hatch" in case they absolutely _want_ Cygwin's default behavior, so I am more in favor of just switching the default and support opting out (via `MSYS=nowincmdln`). Sound okay?

dscho:
I just added a commit to enable `MSYS=wincmdln` by default, and verified that it works.

elieux:
> Having said that, I do want to give users an "escape hatch" in case they absolutely _want_ Cygwin's default behavior, so I am more in favor of just switching the default and support opting out (via `MSYS=nowincmdln`). Sound okay?

Okey dokey.

dscho:
@elieux @lazka so... any chance one of you could review these two patches (and approve)?

elieux:
Sure. But did you add that second commit? I can't see it.

lazka:
> No, this is for MSYS2/Cygwin binaries. If you call, say, `vim.exe abc` from a regular MSYS2 Bash, you won't see the command-line arguments in the output of `wmic process where "CommandLine like '%vim%'" get CommandLine` (see [git-for-windows/git#2756](https://github.com/git-for-windows/git/issues/2756) for more details).

ok, I was confused since the commit says "for native Win32 applications". Thanks for the explanation. With my very limited cpp/cygwin experience, your explanation makes sense, so lgtm.

dscho:
>  did you add that second commit? I can't see it.

I had pushed it, but to another remote. D'oh! Sorry ;-)

dscho:
FWIW [this](https://github.com/msys2/Cygwin/pull/10/commits/fae9acdc62656fbf643dccb716d1e21751335386) is the second commit I was talking about.

lazka (approved):

elieux:
Should the docs be updated or do we not care about them?

elieux:
> ```
> <listitem>
> <para><envar>(no)wincmdln</envar> - if set, the windows complete command
> line (truncated to ~32K) will be passed on any processes that it creates
> in addition to the normal UNIX argv list.  Defaults to not set.</para>
> </listitem>
> ```

elieux:
It is hard for me to follow the code, but comparing our current code and your patched code both to upstream code, yours seems to be more logical. I'm assuming it builds and does what it says on the tin.

elieux (approved):

dscho:
> Should the docs be updated or do we not care about them?

I was on the fence, as the documentation still talks about `CYGWIN` while we ignore that environment variable and look at `MSYS` instead.

> It is hard for me to follow the code, but comparing our current code and your patched code both to upstream code, yours seems to be more logical.

Basically, [@Alexpux' patch](https://github.com/msys2/Cygwin/commit/91990dabdc3c09d1951325d601593cb31815da7c) erroneously moves the `if (wincmdln ...)` block from being a sibling of the `if (real_path.iscygexec ())` block to an `else` arm of the latter. For details, see https://github.com/msys2/Cygwin/commit/91990dabdc3c09d1951325d601593cb31815da7c#diff-b72d3b2289a1e6410a2602d0fa250a3cL386-L395 (and expand the preceding context a bit).

My patch moves the `if (wincmdln ...)` block _out_ of that `else` arm again, so that it also applies to spawned MSYS2 processes.

> I'm assuming it builds and does what it says on the tin.

Yes, I tested (and just re-tested, to be sure) the built `msys0.dll`, verifying that it does what I claim in this PR.

lazka:
offtopic, would it make sense to rename this repo to msys2-runtime? To make the relation to the package more clear?

dscho:
> offtopic, would it make sense to rename this repo to msys2-runtime? To make the relation to the package more clear?

Sure. In Git for Windows, we did exactly that: https://github.com/git-for-windows/msys2-runtime/

lazka:
:+1: done

PR #15, opened 2020-09-22
Fix encoding for console programs heeding `LC_ALL`

### LINKED ISSUE #1974: [BUG] Wrong charset after upgrade to msys2-runtime-3.1.4
## Steps to Reproduce the Problem
`python -c 'print("世界")'`
荳也阜

## Additional Context
`locale`
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_ALL=

## The Following Methods Can All Get Correct Result
1. Downgrade msys2-runtime to 3.0.7
2. `winpty python -c 'print("世界")'`
3. `python -c 'print("世界")' | iconv -f utf-8 -t utf-8`
4. `chcp.com 65001 && python -c 'print("世界")'`

dscho:
Many console programs do not call `SetConsoleOutputCP()` to ensure that the expected code page is active, in particular programs with Linux roots (where you look at `LANG` and `LC_ALL` and `LC_CTYPE`, but not at a code page).

As of Cygwin v3.1.0 (and therefore, MSYS2 v3.1.x), this leads to garbled output, both with and without Pseudo Console support enabled.

Let's work around this.

This fixes https://github.com/msys2/MSYS2-packages/issues/1974

### DIFF
--- a/winsup/cygwin/fhandler_tty.cc
+++ b/winsup/cygwin/fhandler_tty.cc
@@ -2905,6 +2905,8 @@ fhandler_pty_slave::setup_locale (void)
 	  get_ttyp ()->term_code_page = cs_names[i].cp;
 	  break;
 	}
+  SetConsoleCP (get_ttyp ()->term_code_page);
+  SetConsoleOutputCP (get_ttyp ()->term_code_page);
 }
 
 void


dscho:
Note: the upcoming v3.2.0 already has the equivalent patch: 232fde0e76db14fc59912d0022b72c7739f1d598

lazka:
lgtm. This means we'll get the utf-8 code page in the common case, right?

dscho:
> This means we'll get the utf-8 code page in the common case, right?

Yes!

PR #16, opened 2020-10-21
Expose the "deep-copy" symlink behavior via a new `winsymlinks` mode

dscho:
In https://gitter.im/msys2/msys2?at=5f8fe84abbffc02b582d6915, @Alexpux lamented that `tar` does not extract symbolic links but makes deep copies instead.

The reason for this is that we specifically override Cygwin's `symlink()` behavior to make deep copies instead of creating a special-crafted system file that Cygwin (MSYS2, and only those two) will recognize as symbolic link.

The patch to override this behavior is incomplete, though: it hacks up the `WSYM_sysfile` code path, which causes all kinds of problems:

- Contrary to the documentation, `MSYS=winsymlinks:native` won't fall back to creating Cygwin-style symbolic links. Instead, it (quite surprisingly) creates deep copies instead.
- There is no way to opt into the original Cygwin behavior because that code path is now dead code.
- The fact that the `WSYM_sysfile` code path is modified to do something very different from what it is intended to do means that all changes that Cygwin makes to that code path have to be inspected, and occasionally even more hacks have to be put on top, such as https://github.com/msys2/msys2-runtime/commit/44bb1334481a2d7c84588c413f5cace86b8f6e6c, where we specifically had to disable Cygwin's code to try creating WSL symlinks before falling back to creating system files instead).

Let's fix this by introducing a proper new mode: `MSYS=winsymlinks:deepcopy`. This mode is made the default so that users should be unaffected by this PR.

Users wishing to let the MSYS2 runtime create Cygwin-style symbolic links can ask for that by setting `MSYS=winsymlinks:sysfile`.

By fixing this issue, we can now stop disabling Cygwin's WSL symlink behavior, too: it is in the `sysfile` code path and won't affect the default behavior of the MSYS2 runtime.

While at it, the indentation of the original patch is fixed, too.

### DIFF
--- a/winsup/cygwin/environ.cc
+++ b/winsup/cygwin/environ.cc
@@ -87,6 +87,10 @@ set_winsymlinks (const char *buf)
   else if (ascii_strncasematch (buf, "native", 6))
     allow_winsymlinks = ascii_strcasematch (buf + 6, "strict")
 			? WSYM_nativestrict : WSYM_native;
+  else if (ascii_strncasematch (buf, "deepcopy", 8))
+    allow_winsymlinks = WSYM_deepcopy;
+  else
+    allow_winsymlinks = WSYM_sysfile;
 }
 
 /* The structure below is used to set up an array which is used to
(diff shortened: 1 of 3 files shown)


Alexpux (approved):

lazka:
lgtm

elieux:
Looks good. (I didn't build though.)

PR #18, opened 2020-11-08
cygwin: use CREATE_DEFAULT_ERROR_MODE in spawn

jeremyd2019:
This allows subprocesses to get Windows-default error handling behavior (such as invoking the [registered JIT debugger](https://docs.microsoft.com/en-us/windows/win32/debug/configuring-automatic-debugging#configuring-automatic-debugging-for-application-crashes)).

Fixes https://sourceforge.net/p/msys2/discussion/general/thread/1daf3345/ (google turned this thread up when I was looking up how to use `error_start`)

### DIFF
--- a/winsup/cygwin/spawn.cc
+++ b/winsup/cygwin/spawn.cc
@@ -447,6 +447,13 @@ child_info_spawn::worker (const char *prog_arg, const char *const *argv,
 
       c_flags |= CREATE_SEPARATE_WOW_VDM | CREATE_UNICODE_ENVIRONMENT;
 
+      /* Add CREATE_DEFAULT_ERROR_MODE flag for non-Cygwin processes so they
+	 get the default error mode instead of inheriting the mode Cygwin
+	 uses.  This allows things like Windows Error Reporting/JIT debugging
+	 to work with processes launched from a Cygwin shell. */
+      if (!real_path.iscygexec ())
+	c_flags |= CREATE_DEFAULT_ERROR_MODE;
+
       /* We're adding the CREATE_BREAKAWAY_FROM_JOB flag here to workaround
 	 issues with the "Program Compatibility Assistant (PCA) Service".
 	 For some reason, when starting long running sessions from mintty(*),


dscho:
>     * should this be behind an option (like in the CYGWIN/MSYS2 environment variable)?

I think that would be a good idea (note, though, that it is the `MSYS` variable, not the `MSYS2` variable).

The reason I say this is that I would like to have as vanilla an MSYS2 runtime as possible in Git for Windows, where we definitely do _not_ want to invoke the registered debugger.

FWIW it is a bit unclear to me how useful this behavior is in practice: I only know of the Visual Studio debugger as being an option for this, and we do not have any `.pdb` files for Visual Studio to work with.

If it turns out to be quite useful, I would strongly recommend sending this patch to the Cygwin project, too.

jeremyd2019:
> FWIW it is a bit unclear to me how useful this behavior is in practice: I only know of the Visual Studio debugger as being an option for this, and we do not have any `.pdb` files for Visual Studio to work with.

I actually use WinDbg/WinDbgX, now that I've discovered cv2pdb.  But it is possible to use GDB in this way too: https://www.sourceware.org/gdb/current/onlinedocs/gdb/Cygwin-Native.html (not sure why this is documented in a file called 'Cygwin-Native')
> To use it, create or edit the following keys in HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\AeDebug and/or HKLM\SOFTWARE\Wow6432Node\Microsoft\Windows NT\CurrentVersion\AeDebug (for x86_64 versions):
>
> - Debugger (REG_SZ) — a command to launch the debugger. Suggested command is: fully-qualified-path-to-gdb.exe -ex "attach %ld" -ex "signal-event %ld" -ex "continue".
The first %ld will be replaced by the process ID of the crashing process, the second %ld will be replaced by the ID of the event that blocks the crashing process, waiting for GDB to attach.
>
> - Auto (REG_SZ) — either 1 or 0. 1 will make the system run debugger specified by the Debugger key automatically, 0 will cause a dialog box with “OK” and “Cancel” buttons to appear, which allows the user to either terminate the crashing process (OK) or debug it (Cancel).

jeremyd2019:
> > ```
> > * should this be behind an option (like in the CYGWIN/MSYS2 environment variable)?
> > ```
> 
> I think that would be a good idea (note, though, that it is the `MSYS` variable, not the `MSYS2` variable).

Great, now I have to come up with a name for this 😬

jeremyd2019:
Best I could come up with for a name was `spawndflterrmode` but I am definitely open to suggestions.  The more I look at that the less I like it 🤮

dscho:
Why not `jitdebugging`?

dscho:
> it is possible to use GDB in this way too: https://www.sourceware.org/gdb/current/onlinedocs/gdb/Cygwin-Native.html (not sure why this is documented in a file called 'Cygwin-Native')

With that as supporting evidence, it should be easy to "sell" this patch upstream, too.

jeremyd2019:
> Why not `jitdebugging`?

Well, cygwin processes will still require using `error_start`.  Maybe something like `winjitdebug`.  Though the `SEM_NOGPFAULTERRORBOX` flag also seems to suppress Windows Error Reporting.  Unfortunate that so much was 'bolted on' to a flag that was only supposed to suppress a 'General Protection Fault' message on Windows 3.1.

mingwandroid:
I used to use qtcreator as the debugger here when msys2 software crashed along with an msys2 kit in the ide. It worked really well in 2015 odd!

jeremyd2019:
I also think it would be useful to document how to do just-in-time debugging (such as in a 'wiki' page).  For now, I've started a gist with what I've learned: https://gist.github.com/jeremyd2019/3069ae1d44e86f4695087b98271af5d9

dscho:
> I would strongly recommend sending this patch to the Cygwin project, too.

FWIW I still believe that would make sense.

jeremyd2019:
> > I would strongly recommend sending this patch to the Cygwin project, too.
> 
> FWIW I still believe that would make sense.

Just about forgot about this, since I finished with the crash I was trying to catch...
https://cygwin.com/pipermail/cygwin-patches/2020q4/010870.html

jeremyd2019:
Change was applied upstream after some discussion and adjustment: https://github.com/msys2/msys2-runtime/commit/21ec498d7f91285018e74f9767fe46cebace682a

Upstream decided against a new option, and to apply the flag to all non-cygwin processes.

I'm not in a hurry for this change ATM, I'd be content to wait for it to show up in the next stable release, but I cherry-picked the commit and force-pushed prior to closing just in case.

dscho:
> Upstream decided against a new option, and to apply the flag to all non-cygwin processes.

Uh oh. That's _really_ not good, and we only found out _now_: https://github.com/msys2/MSYS2-packages/pull/2414#issuecomment-810841296

jeremyd2019:
🤦 .  I had been using this patch for a while, building clang packages, and I think I saw a debugger pop up once during configure, but no missing dll message boxes.

dscho:
You probably did not try to run any `.exe` with a missing `.dll` ;-)

PR #21, opened 2020-12-13
bug fix: ldd silent about missing dll dependencies

jannick0:
As of now `ldd` is silent about missing dll dependencies - and it always returns with `0`, regardless if all dll dependencies are found or not.

With the commits kept as minimal as possible, the PR fixes:
* missing dll dependencies are now listed marked with `not found` (dll dependencies on a dll which is not found are still not listed, of course)
* `ldd` returns with exit code `1` if not all dll dependencies are found.
* memory leakage as an extra

This bug is a bit painful per se - and, in particular, for the CI function `list_dll_deps` in `ci-library.sh` of the MSYS and MINGW repos:
~~~sh
# from ci-library.sh
# List DDL dependencies
list_dll_deps(){
    local target="${1}"
    echo "$(tput setaf 2)MSYS2 DLL dependencies:$(tput sgr0)"
    find "$target" -regex ".*\.\(exe\|dll\)" -print0 | xargs -0 -r ldd | GREP_COLOR="1;35" grep --color=always "msys-.*\|" \
    || echo "        None"
}
~~~
1. `list_dll_deps` could be amended, e.g.,
* for coloring and
*  such that CI fails or sends out an email if a dll dependency cannot be resolved (e.g., due to a changed dll file name)
2. The bug persists on `Cygwin`.  In case one of you guys think this should be patched upstream, please go ahead.  I am not really involved with the cygwin infrastructure I must admit.  Thanks.

I hope this PR sits on the correct branch of the repo. This is my first PR here.

### DIFF
--- a/winsup/utils/ldd.cc
+++ b/winsup/utils/ldd.cc
@@ -407,6 +407,13 @@ report (const char *in_fn, bool multiple)
 	  }
 	  break;
 	case EXIT_PROCESS_DEBUG_EVENT:
+    if (ev.u.ExitProcess.dwExitCode != 0)
+    {
+      /* All dll dependencies were NOT found. */
+      process_fn = fn_win;
+      res = 1;
+    }
+
 print_and_exit:
 	  print_dlls (&dll_list, isdll ? fn_win : NULL, process_fn);
 	  exitnow = true;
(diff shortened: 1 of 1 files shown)


dscho:
Is this maybe addressing bugs in Cygwin directly?

jannick0:
Yes - see 2.) above.

dscho:
> In case one of you guys think this should be patched upstream, please go ahead.

I won't have the time (and even if I had the time, I don't like to be treated like a convenience, so I'll decline).

PR #25, opened 2021-01-26
Add serial port MARK and SPACE parity support

maras420:


Biswa96:
Should not this be in upstream cygwin repository first?

lazka:
yes, non MSYS2 specific patches need to go to cygwin

lazka:
@maras420 saw your commit, thanks for upstreaming!

PR #27, opened 2021-03-12
Treat "app execution aliases" as symlinks, and allow executing them

### LINKED ISSUE #1943: [BUG] MSYS2 fails to access Microsoft Store-distributed Python 3
## Describe the bug

MSYS shell fails to run Python 3 with "Permission denied" error.

## Steps to Reproduce the Problem

1. Install MSYS2 from https://www.msys2.org/
2. Install Python 3 from https://www.microsoft.com/store/productId/9MSSZTT1N39L
3. Run `/c/Users/Kagami/AppData/Local/Microsoft/WindowsApps/python3.exe` from MSYS2

Expected behavior: It should run Python 3

## Additional Context: Operating System, Screenshots

- OS: Windows 10 Pro version 2004 64bit

image

dscho:
This most notably allows running the Windows Store version of Python directly from MSYS2's Bash (without `cmd /c ...` tricks).

This fixes https://github.com/msys2/Msys2-packages/issues/1943

### DIFF
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -2439,6 +2439,22 @@ symlink_info::check_sysfile (HANDLE h)
   return res;
 }
 
+typedef struct _REPARSE_APPEXECLINK_BUFFER
+{
+  DWORD ReparseTag;
+  WORD  ReparseDataLength;
+  WORD  Reserved;
+  struct {
+    DWORD Version;       /* Take member name with a grain of salt. */
+    WCHAR Strings[1];    /* Four serialized, NUL-terminated WCHAR strings:
+			   - Package ID
+			   - Entry Point
+			   - Executable Path
+			   - Application Type
+			   We're only interested in the Executable Path */
+  } AppExecLinkReparseBuffer;
+} REPARSE_APPEXECLINK_BUFFER,*PREPARSE_APPEXECLINK_BUFFER;
+
 static bool
 check_reparse_point_string (PUNICODE_STRING subst)
 {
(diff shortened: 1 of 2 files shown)


dscho:
I submitted these patches to the Cygwin mailing list (cover letter is [here](https://cygwin.com/pipermail/cygwin-patches/2021q1/011250.html), unfortunately I forgot to pass the `--threaded` option to `git format-patch`... _sigh_).

Biswa96:
Are we allowed to use Rtl* functions?

dscho:
> Are we allowed to use Rtl* functions?

Look at the context, e.g. https://github.com/msys2/msys2-runtime/blob/647dff4c7ae94966d96b0a427452d648f13cd2e1/winsup/cygwin/path.cc#L2512. I think we're good.

mingwandroid:
LGTM from 👀

dscho:
For the record, someone raised the concern on the Cygwin mailing list that treating appexeclinks as symlinks might break stuff, e.g. backup/restore software: https://cygwin.com/pipermail/cygwin-patches/2021q1/011253.html

I disagree, as Cygwin already interprets several different constructs as symlinks (old-style sysfiles, native symlinks, WSL symlinks, etc) and backup/restore software would not know how to recreate them faithfully, and therefore I deem the objection to miss the mark.

Nevertheless, that discussion should probably settle before merging this PR, just in case that I am asked to revise (or drop) the patch: I do not want MSYS2 to deviate unnecessarily from Cygwin.

dscho:
I got good feedback from the Cygwin maintainer, sent off v2, and it was applied with two minor modifications as 0631c6644e6398c051dcda1a53c0fc0e40c814a3. So I cherry-picked it, and I think we're good to go now.

dscho:
@lazka what should we do with this? Merge, even if v3.2.0 might be just around the corner?

lazka:
I'd wait for 3.2.0 if you don't mind

PR #33, opened 2021-03-22
uname: limit sysname to "MSYS" or "MINGW"

lazka:
uname just included the content of the MSYSTEM env var as sysname
which leads to problems in case build tools use it detect a mingw build
and MSYSTEM is something like UCRT or CLANG.

To work around that, in case MSYSTEM is not set, empty, or not MSYS, then
we just output a hardcoded "MINGW".

### DIFF
--- a/winsup/cygwin/uname.cc
+++ b/winsup/cygwin/uname.cc
@@ -37,9 +37,11 @@ uname_x (struct utsname *name)
       memset (name, 0, sizeof (*name));
       /* sysname */
       char* msystem = getenv("MSYSTEM");
-      const char *msystem_msys = "MSYS";
+      const char* msystem_sysname = "MSYS";
+      if (msystem != NULL && *msystem && strcmp(msystem, "MSYS") != 0)
+        msystem_sysname = "MINGW";
       __small_sprintf (name->sysname, "%s_%s-%u%s",
-		       msystem ? msystem : msystem_msys,
+		       msystem_sysname,
 		       wincap.osname (), wincap.build_number (),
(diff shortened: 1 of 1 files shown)


winsup/cygwin/uname.cc
```diff
@@ -37,9 +37,11 @@ uname_x (struct utsname *name)
       memset (name, 0, sizeof (*name));
       /* sysname */
       char* msystem = getenv("MSYSTEM");
-      const char *msystem_msys = "MSYS";
+      const char* msystem_sysname = "MSYS";
+      if (msystem != NULL && strlen(msystem) != 0 && strcmp(msystem, "MSYS") != 0)
```
dscho: This is a more canonical test for a non-empty string:

```suggestion
      if (msystem != NULL && *msystem && strcmp(msystem, "MSYS") != 0)
```

dscho (approved):
Just a minor suggestion. I'll not insist on using it ;-)

jeremyd2019:
There was at least one example I came across that expected either `MINGW32` or `MINGW64`: msys2/CLANG-packages#15.  I don't know if there are any out there that actually *need* to tell the difference though.

lazka:
> There was at least one example I came across that expected either `MINGW32` or `MINGW64`

good point. we need to check if it actually needs that information. If not I'd rather patch the package.

lazka:
Thanks!

lazka:
Searching around on github I see lots of hardcoded mingw32/64 values, so we need to improve this :/ Also got a report already on discord.

We could check `strstr($MSYSTEM, "32")`. I'm gonna rush this a bit..

PR #35, opened 2021-03-31
Turn off default Windows error reporting by default

dscho:
As reported in https://github.com/msys2/MSYS2-packages/pull/2414#issuecomment-810841296, v3.2.0 of the Cygwin runtime has a pretty disruptive new behavior where it shows a message box e.g. when a DLL is missing (it used to write that error message to `stderr` instead).

Let's reinstate the old behavior, allowing to opt-in to the new behavior (which is desired e.g. when debugging using a registered JIT debugger) via the `winjitdebug` flag in the environment variable `MSYS`.

This implements what we talked about [here](https://github.com/msys2/msys2-runtime/pull/18#issuecomment-724880792), at long last.

Cc: @lazka @jeremyd2019


dscho:
I just verified that this fix works.

lazka:
lgtm

PR #38, opened 2021-04-29
Cygwin: path_conv: Try to handle native symlinks more sanely

orgads:
For local paths, add a check if the inner path components contain native
symlinks or junctions.  Compare the incoming path with the path returned
by NtQueryInformationFile(FileNameInformation).  If they differ, there
must be at least one native symlink or junction in the path.  If so,
treat the currently evaluated file as non-existant.  This forces
path_conv::check to backtrack inner path components until we eliminated
all native symlinks or junctions and have a normalized path.

(cherry picked from commits 456c3a46386f38887407603b2c64b7f63a4871c5 and
13fd26ecf5ca8417146d57b45aed0133435c3497)

### DIFF
--- a/winsup/cygwin/dcrt0.cc
+++ b/winsup/cygwin/dcrt0.cc
@@ -742,6 +742,12 @@ init_windows_system_directory ()
 	  system_wow64_directory[system_wow64_directory_length++] = L'\\';
 	  system_wow64_directory[system_wow64_directory_length] = L'\0';
 	}
+      /* We need the Windows dir in path.cc. */
+      wcscpy (windows_directory, windows_system_directory);
+      windows_directory_length = windows_system_directory_length - 1;
+      windows_directory[windows_directory_length] = L'\0';
+      while (windows_directory[windows_directory_length - 1] != L'\\')
+	windows_directory[--windows_directory_length] = L'\0';
 #endif /* __i386__ */
     }
 }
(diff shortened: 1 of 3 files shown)


lazka:
Thanks

jeremyd2019:
I think this has broken the ability to work with `subst`ed drives
```
$ subst
T:\: -> C:\blah\blah\MINGW-packages

$ ls /t/
ls: cannot access '/t/': No such file or directory
```

This worked fine with 3.2.0-3

orgads:
Thanks for the report. I passed it on to [cygwin mailing list](https://cygwin.com/pipermail/cygwin/2021-May/248437.html).

Ah, now I see you already did :)

jeremyd2019:
How important is this patch?  I use subst all the time for building packages, as a way to shorten paths, so this is pretty much a showstopper for me (to keep me from updating msys2-runtime until this is fixed).

jeremyd2019:
Well, I tried to but I never saw it show up in the archives so assumed it went to the black hole.  But wait, figured out my problem, I was still on the April page 😊

orgads:
I needed it mostly to make cmake tests pass. Let's give them 2-3 days to reply. If they don't, we can revert it.

orgads:
Notice that they usually *don't* CC the author on reply, so if you haven't subscribed to the mailing list, you won't get notified. You need to open the archives once in a while to see if you have a reply.

jeremyd2019:
That was my problem, I had the archives page open in a tab to periodically refresh and see if anything interesting to me showed up, but I didn't notice I was still on the April page now that it's May

Issue #41, opened 2021-05-05
Investigate symlink patch breaking subst paths


lazka:
See https://github.com/msys2/msys2-runtime/pull/40 -> https://github.com/msys2/MSYS2-packages/issues/2471#issuecomment-832157860

Upstream questions:

* https://cygwin.com/pipermail/cygwin/2021-May/248436.html
* https://cygwin.com/pipermail/cygwin/2021-May/248437.html

jeremyd2019:
> Compare incoming path with path returned by NtQueryInformationFile(FileNameInformation). If they differ, bail out as if the file doesn't exist.  This forces path_conv::check to backtrack inner path components.

This works if the reason they differ is because there were symlinks involved somewhere in the path, but does not work if there's some other reason (WoW fs redirection and subst have been revealed so far).  Not so sure that that was a safe assumption to make

jeremyd2019:
https://cygwin.com/pipermail/cygwin/2021-May/248457.html

So now the question becomes, what's different between cygwin and msys2-runtime that would cause this...

lazka:
fstab setup makes the difference: https://cygwin.com/pipermail/cygwin/2021-May/248458.html

jeremyd2019:
https://cygwin.com/pipermail/cygwin/2021-May/248466.html
https://github.com/msys2/msys2-runtime/commit/19d59ce75d5301ae167b421111d77615eb307aa7

I wonder if treating subst paths as symlinks will cause issues for me using them as a way to get shorter paths for builds.  Seems like things could silently 'dereference' them back to their too-long variants.

lazka:
Just tried and cygpath de-references the symlink when converting to a Windows path.

jeremyd2019:
☹️

lazka:
fixed https://github.com/msys2/msys2-runtime/pull/42#issuecomment-843520385

jeremyd2019:
It appears this was only fixed for the *root* of a SUBST drive... https://cygwin.com/pipermail/cygwin/2021-May/248612.html

jeremyd2019:
https://cygwin.com/pipermail/cygwin-patches/2021q2/011405.html

That turned out not to help, as `makepkg` calls `pwd -P` which intentionally dereferences symlinks.

jeremyd2019:
[...]

PR #42, opened 2021-05-09
Reapply Cygwin: path_conv: Try to handle native symlinks more sanely

orgads:
For local paths, add a check if the inner path components contain native
symlinks or junctions.  Compare the incoming path with the path returned
by NtQueryInformationFile(FileNameInformation).  If they differ, there
must be at least one native symlink or junction in the path.  If so,
treat the currently evaluated file as non-existant.  This forces
path_conv::check to backtrack inner path components until we eliminated
all native symlinks or junctions and have a normalized path.

(cherry picked from commits 456c3a46386f38887407603b2c64b7f63a4871c5,
13fd26ecf5ca8417146d57b45aed0133435c3497 and
19d59ce75d5301ae167b421111d77615eb307aa7)

### DIFF
--- a/winsup/cygwin/dcrt0.cc
+++ b/winsup/cygwin/dcrt0.cc
@@ -742,6 +742,12 @@ init_windows_system_directory ()
 	  system_wow64_directory[system_wow64_directory_length++] = L'\\';
 	  system_wow64_directory[system_wow64_directory_length] = L'\0';
 	}
+      /* We need the Windows dir in path.cc. */
+      wcscpy (windows_directory, windows_system_directory);
+      windows_directory_length = windows_system_directory_length - 1;
+      windows_directory[windows_directory_length] = L'\0';
+      while (windows_directory[windows_directory_length - 1] != L'\\')
+	windows_directory[--windows_directory_length] = L'\0';
 #endif /* __i386__ */
     }
 }
--- a/winsup/cygwin/globals.cc
+++ b/winsup/cygwin/globals.cc
@@ -26,6 +26,8 @@ UINT windows_system_directory_length;
 #ifdef __i386__
 WCHAR system_wow64_directory[MAX_PATH];
 UINT system_wow64_directory_length;
+WCHAR windows_directory[MAX_PATH];
+UINT windows_directory_length;
 #endif /* __i386__ */
 WCHAR global_progname[NT_MAX_PATH];
 
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1017,17 +1017,25 @@ path_conv::check (const char *src, unsigned opt,
 		    }
 		  goto out;	// file found
 		}
-	      /* Found a symlink if symlen > 0.  If component == 0, then the
-		 src path itself was a symlink.  If !follow_mode then
-		 we're done.  Otherwise we have to insert the path found
-		 into the full path that we are building and perform all of
-		 these operations again on the newly derived path. */
-	      else if (symlen > 0)
+	      /* Found a symlink if symlen > 0 or short-circuited a native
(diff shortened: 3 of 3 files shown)


orgads:
Now with the subst fix.

jeremyd2019:
#41

I am concerned that the subst fix of treating subst drives as symlinks will result in automatically 'dereferencing' the subst when running native Windows processes, undoing the attempted fix for too-long paths used by some build tools (which encode the *ENTIRE PATH TO THE ROOT* in the object file name 🤦)  I will test this, as well as a simple tweak:
```diff
diff --git a/winsup/cygwin/path.cc b/winsup/cygwin/path.cc
index 4ebaf4dc6..53534b8b6 100644
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -3670,8 +3670,7 @@ restart:
 		   somewhere else, thus, it's a symlink in POSIX speak. */
 		if (upath.Length == 14)	/* \??\X:\ */
 		  {
-		    fileattr &= ~FILE_ATTRIBUTE_DIRECTORY;
-		    path_flags |= PATH_SYMLINK;
+		    goto file_not_symlink;
 		  }
 		/* For final paths differing in inner path components return
 		   length as negative value.  This informs path_conv::check
```

jeremyd2019:
3.2.0-6
```
$ cd /t && cmd /c cd
T:\
```

dll from this PR
```
$ cd /t && cmd /c cd
E:\MINGW-packages
```

dll from this PR with above patch applied:
```
$ cd /t && cmd /c cd
T:\
```

orgads:
Is it bad? How does cmd behave?

jeremyd2019:
I don't think it would normally be bad, however, there is a known issue when building packages from MINGW-packages and/or MSYS2-packages that, if the full path is too long the package build will fail.  And "too long" is not really all that long at all.

The CI scripts here work around this by moving the checkout to paths like `C:\_`.
https://github.com/msys2/MINGW-packages/actions/runs/830099565/workflow#L51

I work around it by using subst to shorten the paths.
https://github.com/jeremyd2019/msys2-build32/actions/runs/828859516/workflow#L96

If this code transparently dereferences the subst, as it seems to do, that defeats the purpose of the subst for me, and I'd probably have to switch to moving my checkout instead.  Not the end of the world but breaking my workflows (I also do this subst trick on my local machines for the same reason).

jeremyd2019:
https://cygwin.com/pipermail/cygwin/2021-May/248521.html

orgads:
I see your point. Please refer to [my comment](https://github.com/msys2/setup-msys2/issues/136#issuecomment-842599390).

How do you suggest to proceed?

jeremyd2019:
I have sent my quick and dirty patch as a reply to the "all-or-nothing" comment.  I don't know if I can make a sufficiently compelling argument about cygwin catering to brain-dead tools versus being more theoretically correct.  But since msys2's *raison d'être* is running these build tools, it may be something that this fork should work around (such as with my patch above)

jeremyd2019:
My patch was applied upstream!  https://github.com/msys2/msys2-runtime/commit/c8949d04001e3dbc03651475b6cd1c5623400835

lazka:
nice, backported. I guess this is OK now?

orgads:
Fine with me. Maybe squash the commits?

jeremyd2019:
I was about to indicate that it looks OK to me now, except I noticed the fs redirection assumption within `#ifdef __i386__` and thought about the x86_64-on-arm64 case.   But I just tested it and I didn't notice any issues ls-ing system32, so :+1:

lazka:
> Fine with me. Maybe squash the commits?

We will clean up the history on the next cygwin release. This makes it clearer that it can be dropped.

PR #46, opened 2021-06-04
make new native inner symlink code respect PC_SYM_FOLLOW and PC_SYM_NOFOLLOW_REP flags, and add option to disable it

### LINKED ISSUE #41: Investigate symlink patch breaking subst paths

jeremyd2019:
I've cherry-picked the commits that made it upstream, plus added an option "nativeinnerlinks" which is on by default but can disable this new code for cases where it causes problems (most likely with SUBST and mapped network drives), by setting `MSYS=nonativeinnerlinks`.  This patch, and the option, fixes #41.

### DIFF
--- a/winsup/cygwin/environ.cc
+++ b/winsup/cygwin/environ.cc
@@ -127,6 +127,7 @@ static struct parse_thing
   {"disable_pcon", {&disable_pcon}, setbool, NULL, {{false}, {true}}},
   {"enable_pcon", {&disable_pcon}, setnegbool, NULL, {{true}, {false}}},
   {"winjitdebug", {&winjitdebug}, setbool, NULL, {{false}, {true}}},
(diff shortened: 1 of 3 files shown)


jeremyd2019:
I've cherry-picked the commits that made it upstream here.  As I mentioned in #41, this is not sufficient to deal with the issues I've had.

jeremyd2019:
Any objections to this (adding option to disable this new code)?

jeremyd2019:
Hearing none...

PR #50, opened 2021-06-22
Don't fall back to ExitProcess if process handled Ctrl-C event

### LINKED ISSUE #49: Ctrl-C handling: don't fall back to ExitProcess if process handled Ctrl-C event
> > should it force the issue with ExitProcess at all?
> 
> Hmm. I think your Python example points to the answer "no"... I had not thought about that.

> Or somehow tell if the console control event was handled or not? (if that's possible)

I _guess_ that the return value of `CtrlRoutine` would tell, but I am not sure. Since this function is undocumented, it will require some experimentation, methinks.

_Originally posted by @dscho in https://github.com/msys2/msys2-runtime/issues/48#issuecomment-865346513_

naveen521kk:
I have experimented with the mysterious kernel32!CtrlRoutine that
it returns 0 if the CtrlHandlers set using SetConsoleCtrlHandler
handled it correctly. So, now rather than defaulting to move to
ExitProcess if the process isn't killed by CtrlRoutine, it will
check only if the ctrl event is handled correctly and falls
back to ExitProcess only when CtrlRoutine fails.

This should possibly allow python interactive mode.

Fixes https://github.com/msys2/msys2-runtime/issues/49

/cc @dscho @jeremyd2019

### DIFF
--- a/winsup/utils/getprocaddr.c
+++ b/winsup/utils/getprocaddr.c
@@ -40,12 +40,23 @@ inject_remote_thread_into_process (HANDLE process,
     {
       /*
        * Wait up to 10 seconds (arbitrary constant) for the thread to finish;
-       * After that grace period, fall back to exit with an exit code
-       * indicating failure.
+       * Maybe we should wait forever? I have seen Cmd does so, but well...
        */
-      if (WaitForSingleObject (thread, 10000) == WAIT_OBJECT_0 &&
-          GetExitCodeThread(thread, &code) && code != STILL_ACTIVE)
+      if (WaitForSingleObject (thread, 10000) == WAIT_OBJECT_0)
         res = 0;
+      /*
+      According to the docs at MSDN for GetExitCodeThread, it will
+      get the return value from the function, here CtrlRoutine. So, this
+      checks if the Ctrl Event is handled correctly by the process.
+
+      By some testing I could see CtrlRoutine returns 0 in case where
+      CtrlEvent set by SetConsoleCtrlHandler is handled correctly, in all
+      other cases it returns something non-zero(not sure what it that).
+      */
+      GetExitCodeThread (thread, &code);
+      if (code != 0)
+        res = code;
+
       CloseHandle (thread);
     }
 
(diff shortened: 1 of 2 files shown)


naveen521kk:
The small test program I used was
```c
#include <windows.h>
BOOL control_handler(DWORD control_type) {
    return FALSE;
}
int main(){
    SetConsoleCtrlHandler((PHANDLER_ROUTINE) control_handler, TRUE);
    Sleep(600000);
}
```
When `control_handler` returned `TRUE`, `GetExitCodeThread` returned `0`. When `control_handler` returned `FALSE`, it was something else but not Zero.

Here, [SetConsoleCtrlHandler](https://docs.microsoft.com/en-us/windows/console/handlerroutine#return-value) says if `control_handler` returns false means it failed to handle the control event. I guess this should allow Python to run interactively.

dscho (approved):
Yay!

naveen521kk:
I think there is some problem with python though,
```py
$ /c/msys64/mingw64/bin/python.exe
Python 3.8.9 (default, Apr 13 2021, 15:54:59)  [GCC 10.2.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> print("hello")
hello
>>>
  File "<stdin>", line 0

    ^
SyntaxError: decoding with 'cp1252' codec failed (KeyboardInterrupt: )
```
after I press Ctrl+C some other things are happening, but I think that isn't related to this patch and would probably need fix in python itself.

dscho:
> ```python
> >>>
>   File "<stdin>", line 0
> 
>     ^
> SyntaxError: decoding with 'cp1252' codec failed (KeyboardInterrupt: )
> ```

I _guess_ that we're not "swallowing" the <kbd>Ctrl</kbd>+<kbd>C</kbd> "character" appropriately (which would be a 0x03 IIRC)? The culprit for this symptom (as for so many other symptoms) could be the ConPTY code, which changed behavior in so many ways even when turning off ConPTY... At least that's my guess, based on the experience of the past year.

dscho:
Just like with @jeremyd2019's patches, I plan on squashing this patch into the appropriate commit(s) during the next rebase.

Thank you, @naveen521kk and @jeremyd2019 for making the <kbd>Ctrl</kbd>+<kbd>C</kbd> handling in the MSYS2 runtime awesome!

winsup/utils/getprocaddr.c
```diff
@@ -40,12 +40,23 @@ inject_remote_thread_into_process (HANDLE process,
...
+      get the return value from the function, here CtrlRoutine. So, this
+      checks if the Ctrl Event is handled correctly by the process.
+
+      By some testing I could see CtrlRoutine returns 0 in case where
+      CtrlEvent set by SetConsoleCtrlHandler is handled correctly, in all
+      other cases it returns something non-zero(not sure what it that).
+      */
+      GetExitCodeThread (thread, &code);
+      if (code != 0)
+        res = code;
```
jeremyd2019, line 58: It seems a little weird to conflate an error injecting the thread with the return value from the injected thread.  But I guess it works...

PR #52, opened 2021-06-23
getprocaddr: refactor/cleanup

### LINKED ISSUE #51: Try finding CtrlRoutine with GetProcAddress
Oddly enough, this hack seems to work:
```diff
diff --git a/winsup/utils/getprocaddr.c b/winsup/utils/getprocaddr.c
index 80b399c..beafc10 100644
--- a/winsup/utils/getprocaddr.c
+++ b/winsup/utils/getprocaddr.c
@@ -164,9 +164,9 @@ main (int argc, char **argv)
    * because here because that symbol isn't exported.
    */

-  if (strcmp (argv[1], "CtrlRoutine"))
+  if (TRUE || strcmp (argv[1], "CtrlRoutine"))
     {
-      HINSTANCE kernel32 = GetModuleHandle ("kernel32");
+      HINSTANCE kernel32 = GetModuleHandle (strcmp(argv[1], "CtrlRoutine") ? "kernel32" : "kernelbase");
       if (!kernel32)
         return 1;
       void *address = (void *)GetProcAddress (kernel32, argv[1]);
```

Maybe it could look in `KERNELBASE` for `CtrlRoutine` the same way it looks in `KERNEL32` for ExitProcess, and if it's not found fall back to using dbghelp (assuming there's a downlevel version of Windows for which that didn't work, that prompted the dbghelp approach in the first place).

_Originally posted by @jeremyd2019 in https://github.com/msys2/msys2-runtime/issues/47#issuecomment-865556700_

jeremyd2019:
Look in kernel32 and kernelbase for `CtrlRoutine` before breaking out the
big guns and trying to find it in the call stack.  Fixes #51.

The change to look at the return code of the thread caused the program
to act as though ExitProcess failed to inject.  Deal with this by
passing the thread exit code as an OUT param, and only looking at it
when calling the CtrlRoutine.

While I was there, I noticed a potential race between `CreateEvent` and
the `ctrl_handler` function that uses it, so cleaned that up.  Also the
event doesn't need to be named, so don't name it to avoid potential
collisions with other processes.

Fixed a couple of printf type warnings.  As this is mingw, it should be
safe to use `%lu` for DWORDs.

### DIFF
--- a/winsup/utils/getprocaddr.c
+++ b/winsup/utils/getprocaddr.c
@@ -27,13 +27,14 @@ static HANDLE CtrlEvent;
 static int
 inject_remote_thread_into_process (HANDLE process,
                                    LPTHREAD_START_ROUTINE address,
-                                   uintptr_t exit_code)
+                                   uintptr_t exit_code,
+                                   DWORD *thread_return)
 {
   int res = -1;
 
   if (!address)
     return res;
-  DWORD thread_id, code;
+  DWORD thread_id;
   HANDLE thread = CreateRemoteThread (process, NULL, 1024 * 1024, address,
                                       (PVOID)exit_code, 0, &thread_id);
   if (thread)
@@ -53,9 +54,8 @@ inject_remote_thread_into_process (HANDLE process,
       CtrlEvent set by SetConsoleCtrlHandler is handled correctly, in all
       other cases it returns something non-zero(not sure what it that).
       */
-      GetExitCodeThread (thread, &code);
-      if (code != 0)
-        res = code;
+      if (thread_return != NULL)
+        GetExitCodeThread (thread, thread_return);
 
       CloseHandle (thread);
     }
@@ -80,6 +80,7 @@ ctrl_handler (DWORD ctrl_type)
   HANDLE process;
   PSYMBOL_INFOW info;
   DWORD64 displacement;
+  DWORD thread_return = 0;
 
   count = CaptureStackBackTrace (1l /* skip this function */,
                                  1l /* return only one trace item */, &address,
@@ -134,15 +135,18 @@ ctrl_handler (DWORD ctrl_type)
           return 1;
         }
       /* Inject the remote thread only when asked to */
-      int t = inject_remote_thread_into_process (h, address, exit_code);
-      if (t != 0)
+      if (inject_remote_thread_into_process (h, address, exit_code,
+                                             &thread_return) < 0)
         {
           fprintf (stderr,
-                   "Error while injecting remote thread %d for pid(%d)\n", t,
-                   pid);
+                   "Error while injecting remote thread for pid(%lu)\n", pid);
           exit (1); /*We should exit immediately or else there will a 10s hang
                        waiting for the event to happen.*/
         }
+      if (thread_return)
+        fprintf (stderr,
+                 "Injected remote thread for pid(%lu) returned %lu\n", pid,
+                 thread_return);
     }
   SymCleanup (process);
   if (!SetEvent (CtrlEvent))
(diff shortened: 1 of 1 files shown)


jeremyd2019:
Arm binaries as of 4afb789:
[getprocaddrarm.zip](https://github.com/msys2/msys2-runtime/files/6699234/getprocaddrarm.zip)

jeremyd2019:
I've only tested this on Win7 so far.

winsup/utils/getprocaddr.c
```diff
@@ -150,77 +151,16 @@ ctrl_handler (DWORD ctrl_type)
...
 }
 
-int
-main (int argc, char **argv)
+/* The easy route for finding the address of CtrlRoutine
+ * would be use GetProcAddress() but this isn't viable
+ * here because that symbol isn't exported.
+ */
+static int
+find_ctrl_routine_the_hard_way ()
```
dscho, line 165: 😁

winsup/utils/getprocaddr.c
```diff
@@ -27,13 +27,14 @@ static HANDLE CtrlEvent;
 static int
 inject_remote_thread_into_process (HANDLE process,
                                    LPTHREAD_START_ROUTINE address,
-                                   uintptr_t exit_code)
+                                   uintptr_t exit_code,
+                                   DWORD * thread_return)
```
dscho: Isn't Cygwin's code style cuddling the asterisk with the variable name?

dscho (approved):
I like it! Especially if that allows us to avoid getting the `CtrlRoutine` address "the hard way".

winsup/utils/getprocaddr.c
```diff
@@ -134,11 +135,11 @@ ctrl_handler (DWORD ctrl_type)
           return 1;
         }
       /* Inject the remote thread only when asked to */
-      int t = inject_remote_thread_into_process (h, address, exit_code);
+      int t = inject_remote_thread_into_process (h, address, exit_code, &thread_return);
```
naveen521kk: Do you need this variable `t`? The return will be either `0` or `-1` so rather than assigning to `t` it can be directly checked in an `if` condition.

Also, I would like `thread_return` to be printed.

winsup/utils/getprocaddr.c
jeremyd2019: Damned if I know.  I'll change it though

winsup/utils/getprocaddr.c
```diff
@@ -249,32 +192,119 @@ main (int argc, char **argv)
...
+  if (inject_remote_thread_into_process (h, (LPTHREAD_START_ROUTINE)address,
+                                         exit_code, &thread_return) < 0)
+    {
+      fprintf (stderr, "Could not inject thread into process %lu\n", pid);
       return 1;
     }
+
+  if (is_ctrl_routine && thread_return)
+    {
+      fprintf (stderr,
```
dscho, line 303: I don't really know whether it makes sense to have this `fprintf()` (and the likewise new one in `ctrl_handler()`). In the regular case, it will go to a hidden console. I guess it does not hurt... But does it benefit anyone?

jeremyd2019: @naveen521kk ? https://github.com/msys2/msys2-runtime/pull/52#discussion_r656997079

dscho: I guess I can live with it as-is ;-)

jeremyd2019: I would still like to hear @naveen521kk's rationale, but it is nice to see when testing.

naveen521kk: While testing the executables through bash I could see it didn't print to hidden console, instead the output was in original console. That's why I wanted it. But yes, this will not work with cmd.

jeremyd2019:
Arm binaries as of 12e06c8:
[getprocaddrarm.zip](https://github.com/msys2/msys2-runtime/files/6711182/getprocaddrarm.zip)

The arm64 works great against clangarm64 python, and arm64 arm32 and 32 work great (manually) injecting <kdb>Ctrl</kbd>+<kdb>C</kbd> against `findstr`.  On released Win10 on Arm64.

PR #56, opened 2021-09-23
reinstate hardlink faking

tomty89:
cygwin had hardlink faking removed with commit 6cabe044dd07c17bea043d924f414d74086df7b9. As opposed to what cygwin does, we allow symlink faking. I'm not seeing a concrete reason that we should not allow that for hardlinks. It would allow msys2 to be used on FAT-family volumes.

Instead of reverting the commit, I simply make it does what symlink_worker() does, i.e. use CopyFile() instead of CopyFileW() and do not call SetFileAttributesW().

Also introduces an environment variable to make hardlink faking always/never occur.

P.S. I haven't run any test on this yet, but please do let me know if I have done anything inappropriate.

### DIFF
--- a/winsup/cygwin/globals.cc
+++ b/winsup/cygwin/globals.cc
@@ -61,6 +61,14 @@ enum winsym_t
   WSYM_deepcopy
 };
 
+/* Should hardlinks be faked.  The value is set depending on the
+   "fakehardlinks" setting of the CYGWIN environment variable. */
+enum fakehard_t {
+  FHARD_never = 0,
+  FHARD_ondemand,
+  FHARD_always
+};
+
 exit_states NO_COPY exit_state;
 
 /* Set in init.cc.  Used to check if Cygwin DLL is dynamically loaded. */
@@ -73,6 +81,7 @@ bool pipe_byte;
 bool reset_com;
 bool wincmdln = true;
 winsym_t allow_winsymlinks = WSYM_deepcopy;
+fakehard_t fake_hardlinks = FHARD_ondemand;
 bool disable_pcon = true;
 bool winjitdebug = false;
 bool nativeinnerlinks = true;
--- a/winsup/cygwin/environ.cc
+++ b/winsup/cygwin/environ.cc
@@ -93,6 +93,19 @@ set_winsymlinks (const char *buf)
     allow_winsymlinks = WSYM_sysfile;
 }
 
+static void
+set_fakehardlinks (const char *buf)
+{
+  if (!buf || !*buf)
+    fake_hardlinks = FHARD_ondemand;
+  else if (ascii_strncasematch (buf, "never", 5))
+    fake_hardlinks = FHARD_never;
+  else if (ascii_strncasematch (buf, "always", 6))
+    fake_hardlinks = FHARD_always;
+  else
+    fake_hardlinks = FHARD_ondemand;
+}
+
 /* The structure below is used to set up an array which is used to
    parse the CYGWIN environment variable or, if enabled, options from
    the registry.  */
@@ -124,6 +137,7 @@ static struct parse_thing
   {"reset_com", {&reset_com}, setbool, NULL, {{false}, {true}}},
   {"wincmdln", {&wincmdln}, setbool, NULL, {{false}, {true}}},
   {"winsymlinks", {func: set_winsymlinks}, isfunc, NULL, {{0}, {0}}},
+  {"fakehardlinks", {func: set_fakehardlinks}, isfunc, NULL, {{0}, {0}}},
   {"disable_pcon", {&disable_pcon}, setbool, NULL, {{false}, {true}}},
   {"enable_pcon", {&disable_pcon}, setnegbool, NULL, {{true}, {false}}},
   {"winjitdebug", {&winjitdebug}, setbool, NULL, {{false}, {true}}},
--- a/winsup/cygwin/fhandler_disk_file.cc
+++ b/winsup/cygwin/fhandler_disk_file.cc
@@ -1198,6 +1198,21 @@ fhandler_disk_file::link (const char *newpath)
       return -1;
     }
 
+  if (fake_hardlinks == FHARD_always)
+    {
+      tmp_pathbuf tp;
+      char *w_newpath;
+      char *w_oldpath;
+      stpcpy (w_newpath = tp.c_get (), newpc.get_win32());
+      stpcpy (w_oldpath = tp.c_get (), pc.get_win32());
+      if (!CopyFile (w_oldpath, w_newpath, TRUE))
+        {
+          __seterrno ();
+          return - 1;
+        }
+      return 0;
+    }
+
   char new_buf[nlen + 5];
   if (!newpc.error)
     {
@@ -1250,8 +1265,27 @@ fhandler_disk_file::link (const char *newpath)
     {
       if (status == STATUS_INVALID_DEVICE_REQUEST
 	  || status == STATUS_NOT_SUPPORTED)
-	/* FS doesn't support hard links.  Linux returns EPERM. */
-	set_errno (EPERM);
+        {
+          if (fake_hardlinks == FHARD_never)
+            {
+              /* FS doesn't support hard links.  Linux returns EPERM. */
+              set_errno (EPERM);
+            }
+          else
+            {
+              tmp_pathbuf tp;
+              char *w_newpath;
+              char *w_oldpath;
+              stpcpy (w_newpath = tp.c_get (), newpc.get_win32());
+              stpcpy (w_oldpath = tp.c_get (), pc.get_win32());
+              if (!CopyFile (w_oldpath, w_newpath, TRUE))
+                {
+                  __seterrno ();
+                  return - 1;
+                }
+              return 0;
+            }
+        }
       else
 	__seterrno_from_nt_status (status);
       return -1;


jeremyd2019:
You want to base this PR on msys2-3_2_0-release branch, not master.  You can hit "edit" up top by the title and it will give a dropdown to change the base branch.  You'll also need to rebase your branch against that branch and force-push

tomty89:
Okay I've just run a build test and run test on my machine. It copies when hard linking fails now when it is on a FAT-family volume. The default behavior on NTFS isn't changed. The env var works as expected as well. In addition to a series of `ln` tests, I also did `pacman -S git`, since the git package is known to have hardlinks. Installation went well now.

I also tested with non-ASCII filenames. It failed with `ENOENT`. However, it is the same case with (fake) symlinking, so I guess this is ready to be merged. (Well, except maybe some code styling problem?)

tomty89:
I've also prepared [a (tested) "wide"-supporting variant](https://github.com/tomty89/msys2-runtime/commit/1653c580db2bea6de9fba892ab18ae36c6d86f53). Let me know if you prefer that over it being "in sync" with symlink faking. (I'm not quite capable of making `recursiveCopy()` "wide"-supporting, I think.)

winsup/cygwin/globals.cc
```diff
@@ -73,6 +81,7 @@ bool pipe_byte;
 bool reset_com;
 bool wincmdln = true;
 winsym_t allow_winsymlinks = WSYM_deepcopy;
+fakehard_t fake_hardlinks = FHARD_ondemand;
```
dscho, line 84: It all looks good to me, except for this backward-incompatible change. There are so many users out there that I suspect a bunch of them relying on the current behavior. I'd therefore suggest to set the default to `FHARD_never` instead.

tomty89: Hmm, but `FHARD_ondemand` does not cause actual behavioral change in the case of NTFS (or any other scenario where hardlinks are supported).

The very reason I proposed this is that msys2 does not exactly work (as in, the installation will be more or less broken) when it is (managed) on filesystems that does not support hardlinks (unless you stay away from packages that have hardlinks -- but then IIRC even coreutils does, just that it won't be broken util there's an update), so I'm not sure what you are referring to with *"relying on the current behavior"*.

It merely avoids e.g. the need of managing an installation on NTFS and use and only use a copy of it on a FAT-family filesystem (i.e. fake hardlinks manually).

dscho: > `FHARD_ondemand` does not cause actual behavioral change in the case of NTFS (or any other scenario where hardlinks are supported).

Indeed. But it does change the behavior where they are _not_ supported. I would expect arcane build setups to be out there that expect to error out in such instances rather than filling up FAT drives with unintentional copies, for example. In other words: this change of behavior would not necessarily be as benign as you try to make me believe it would be.

tomty89: > I would expect arcane build setups to be out there that expect to error out in such instances rather than filling up FAT drives with unintentional copies

To be frank, I'm (still) not seeing the logic / the point you are trying to make here, especially when we fake symlinks by default (and in the symlink case, by nature it can be recursive, a.k.a "deep" copy, which is even less "benign"). Yes it does change the behavior in that case, but the problem is, the current behavior doesn't even promise integrity (which IMHO is also even less "benign").

Btw I found out later that msys2 in general won't fit as well as I expected in exFAT, since with large AU size the required capacity can go up like exponentially (because of the amount of small files). (It kind of loses the point if I turn to a small AU size, AFAIK.) So if this is not a default behavior, I kind of feel like it doesn't mean much (to me) anymore. (But if you like, you can make the change and merge it anyway, of course.)

dscho: You are equating the existing symlink behavior with the hardlink behavior you want to introduce on technical grounds. However, I am telling you that the existing behavior is something users are used to, and the change you want to make is something they are _not_ used to, and given the number of users, it is virtually guaranteed that some users will be surprised in a very bad way.

The problem is not technical. The problem is social, and I did not see any attempt on your side to address that.

tomty89: > The problem is not technical. The problem is social, and I did not see any attempt on your side to address that.

Not really. Rather I'm merely not seeing  how your point on the social aspect could stand, when what you think we should preserve is a behavior that produce nothing but a broken installation. (In other words, it never exactly worked on filesystem that does not support hardlinks anyway, and you think we need to avoid breaking such setup, assuming that it could exist somehow.)

If we were instead in a discussion like, currently we fake hardlinks by default anyway, and I propose that we should use NTFS hardlinks when/where possible, I'd be able to empathize with you if you raise the same kind of social concern. But with what we are discussing here, I'll need some concrete example to see in what scenario it could cause a "social impact/issue", otherwise what you concern would seem vague / purely arbitrary to me.

I'm just really not sure what "relying on it to error out" even means in this particular case. (Like, why would / how could someone need to / have been "rely(ing)" on that anyway?)

There's a difference between "expect" and "rely". "I expect no one would fix this" is not the same as "I rely on no one fixing this". (Let alone whether it always makes sense to suit the latter.)

tomty89: Besides, if the behavior is not "socially problematic" in the case of symlink, I doubt that there can be sensical / strong enough reason for us to prefer behavior preservation over working and consistent behavior. (But of course, that's purely my own stand. I'm sure there are a lot people out there would prefer things to stay broken and wrong as long as that that keeps their setups working.)

dscho: So have you thought about an MSYS2 installation on an NTFS drive that is used on a FAT drive?

With my maintainer hat on, I really do not want to see such behavioral change, certainly when unintended consequences are not even considered fully.

tomty89: > So have you thought about an MSYS2 installation on an NTFS drive that is used on a FAT drive?

Uhm yes? As I said, that's what one can / should do until this is merged, and the copy will be have all hardlinks "dereferenced". And unless this is merged, one cannot exactly "manage" (i.e. update) the copy inplace, but will have to keep copying (well, or rsync).

Am I really the one missing something here or is it actually you? (Like, maybe you thought Windows will skip all the hardlinks when one perform such copying?)

dscho: Correct me if I am wrong. Right now, trying to hardlink on FAT will trigger an EPERM. With your patch, it will not do that but quietly copy instead (unless specifically overridden), right up until the disk is full.

To you, this won't be an unexpected change of behavior. To everybody else, it will be.

tomty89: > Right now, trying to hardlink on FAT will trigger an EPERM.

Yes, but only if you try to hardlink explicitly. Copying an installation with hardlinks from an NTFS volume will not trigger any error or even a prompt unless I remember wrongly (at least not with the normal GUI copy and paste). All the hardlinks will be made copies in the FAT volume.

> You are equating the existing symlink behavior with the hardlink behavior you want to introduce on technical grounds.

Or are you trying to be purely technical now? As you are still not giving any scenario that making it no longer EPERM will be "socially" considered "wrong" instead of just "unexpected" (as in, "didn't expect this finally got fixed").

The only reason I relate what we do for symlink to this is that it is the base / reference for the social aspect. Since we agreed / accepted that there could be "copies filling up the drive" when we fake symlink, there's hardly a reason that we cannot accept that when we have to fake hardlinks.

> and I did not see any attempt on your side to address that.

To be frank, if this is all about "I don't want anything to change no matter if it makes sense", you can simply just say it. There's no point sounding it as if I or anyone could ever "address" anything.

tomty89: Or if this is about "people should be aware that having msys2 on FAT volume instead of NTFS volume will additionally require hardlinks to be faked", and *you* think we should rely on the EPERM for that (instead of having a note on some README or so), there's nothing for me to address either, since it's your personal preference (and I have no intention to persuade you or anyone on that).

And ironically, one will only realize it when he/she installs/updates a package with pacman. I doubt that he/she will have any problem extracting a new installation with any of the available package / approach, as the hardlinks will be dereferenced / were not even preserved.

dscho: Look, I've been maintaining software for a long, long time, longer than Git exists. One develops a sense of what poses maintenance problems over the course of such a long time.

And changing the behavior in a way as you are trying to do is such a problem. Of course, it is not _your_ problem because you simply go away after this PR is merged. The onus is firmly on others' shoulders. In this instance, among the few others who maintain the MSYS2 runtime: me.

So I won't merge this PR as-is.

Having said that, if you change the default behavior back to the current behavior, I am inclined to merge it.

tomty89: As I said, I don't even care if this will be merged anymore, because at the the end of the day I realized I can't exactly get any advantage by having msys2 on (ex)FAT without paying some unreasonable price (the AU size problem).

The only reason I continued the discussion is that you sounded like you missed something and/or there's room for discussion.

Although my experience is not as old as git, I've seen enough maintainers / developers who are allergic to *any* change (even when it's logical / reasonable / actually a bug fix).

And you know what, that's totally alright, what's not is that some incline to make it sound like who proposed the change / fix is "not seeing the bigger picture" / "missing some of the caveats", without providing what exactly those can be (because he/she simply couldn't). And apparently, you are yet another one of them.

If all you wanted to say is "I just don't like it", please just say it at the very beginning instead of wasting both of our time.

dscho: > If all you wanted to say is "I just don't like it", please just say it at the very beginning instead of wasting both of our time.

That's a mischaracterization of what I said. You spent my time, too, you know? I think the new option makes sense. But it does not make sense to change the default behavior, which you seem to be unwilling to accept.

tomty89: Btw, I have no intention to be credited for some decision that I can't even tell / am not even told why. I can't / won't stop you from making any further change you want and merge it though, if you have such desire.

tomty89: > That's a mischaracterization of what I said.

Anyone actually read what I typed should notice that I tried really hard not to make that characterization by trying to getting just one concrete example / scenario from you on of how / when the behavior could be deemed bad, wrong or even just undesirable by anyone that is being sensible.

Take the Unicode filename issue as an example, you can say that "no one but you will expect that hardlink faking could now work with files with Unicode filename" if I made it work, but that's not a sensible reason for us to avoid such "surprise".

Yes, in this case, copies of files will be made. But there are two reasons why we could / should do this for hardlinks as well:

1. We have been doing this for symlink, in which case even more copies could be made due to the recursive nature.
2. It's not a "plus" (e.g. making use of NTFS hardlink) to have. An installation can end up being broken / incomplete without this "behavioral change".

While I tried (hard) to get an argument / a reason from you that why it still should not be done, all you provided is nothing  but points like "my rich experience tells me that it smells bad" / "someone will be surprised". I really couldn't see how much that differs from "I just don't like it".

Anyway, I'm out of this. It's unfortunate that I was not experienced enough to tell when I simply shouldn't have started a conversation at all.

dscho:
@jeremyd2019 do you want to continue working on this PR?

jeremyd2019:
No

PR #62, opened 2021-10-30
Update to cygwin 3.3.1

lazka:
WIP -> broken

* [ ] doc build broken (was before I think, we just need to skip pdf/html generation now)
* [ ] install broken (something re libmsys-2.0.a)
* [ ] https://github.com/msys2/msys2-runtime/commit/3839f28dbf2d09b98e6c6f37e7e0e46eb3d97dda#diff-4ede10814d0ef87034153c055aab487583a68313ad9a4a3bd1228945ce23e979 needs to be ported
* [ ] maybe patch cleanup?

Fixes #61

PR #63, opened 2021-11-08
Update to Cygwin v3.3.2

### LINKED ISSUE #61: Update to cygwin 3.3.x
https://cygwin.com/git/?p=newlib-cygwin.git;a=shortlog;h=refs/heads/cygwin-3_3-branch

https://cygwin.com/pipermail/cygwin-announce/2021-October/010268.html

https://cygwin.com/pipermail/cygwin-announce/2021-October/010269.html

dscho:
This is the continuation of https://github.com/msys2/msys2-runtime/pull/62, now on top of v3.3.2 instead of v3.3.1.

Auto-generated ChangeLog:
- Add MSYS2 triplet
- Fix msys library name in import libraries
- Rename dll from cygwin to msys
- Add functionality for converting UNIX paths in arguments and environment variables to Windows form for native Win32 applications.
- Add functionality for changing OS name via MSYSTEM environment variables.
- - Move root to /usr. - Change sorting mount points. - By default mount without ACLs. - Can read /etc/fstab with short mount point format.
- Instead of creating Cygwin symlinks, use deep copy by default
- Automatically rewrite TERM=msys to TERM=cygwin With MSys1, it was necessary to set the TERM variable to "msys". To allow for a smooth transition from MSys1 to MSys2, let's simply handle TERM=msys as if the user had not specified TERM at all and wanted us to use our preferred TERM value.
- Do not convert environment for strace
- Special case for converting root directory to have training slash
- dcrt0.cc: Untangle allow_glob from winshell
- dcrt0.cc (globify): Don't quote literal strings differently when dos_spec
- strace.cc: Don't set MSYS=noglob
- Add debugging for build_argv
- Add debugging for strace make_command_line
- environ.cc: New facility/environment variable MSYS2_ENV_CONV_EXCL
- Fix native symbolic link spawn passing wrong arg0
- QueryUnbiasedInterruptTime must be load from kernel32.dll
- strace --quiet: be *really* quiet
- Default to `disable_pcon`
- Introduce the `enable_pcon` value for `MSYS`
- popen: call /usr/bin/sh instead of /bin/sh
- CI: add a GHA for doing a basic build test
- Set up a GitHub Action to keep in sync with Cygwin
- Expose full command-lines to other Win32 processes by default
- Disable the 'cygwin' GitHub workflow
- Do not show Error dialogs by default
- Add a helper to obtain a function's address in kernel32.dll
- Emulate GenerateConsoleCtrlEvent() upon Ctrl+C
- kill: kill Win32 processes more gently
- Cygwin: make option for native inner link handling.
- docs: skip building texinfo and PDF files
- install-libs: depend on the "toollibs"


dscho:
@lazka from my side, this is good to go.

winsup/utils/kill.cc
```diff
@@ -186,10 +187,20 @@ forcekill (pid_t pid, DWORD winpid, int sig, int wait)
...
+      if (DuplicateHandle (cur, h, cur, &h2, PROCESS_CREATE_THREAD |
+			      PROCESS_QUERY_INFORMATION |
+			      PROCESS_VM_OPERATION |
+			      PROCESS_VM_WRITE | PROCESS_VM_READ |
+			      PROCESS_TERMINATE, FALSE, 0))
+      {
+        CloseHandle(h);
+        h = h2;
+      }
+      exit_process_tree (h2, 128 + sig);
```
jeremyd2019: Would it be better to do
```suggestion
      exit_process_tree (h, 128 + sig);
```
in case DuplicateHandle failed?  I *think* h2 might be uninitialized in that case?

dscho: Right, and the motivation was clearly to use `h` in any case. I _guess_ we can slip this into the PR, although it would make sense to backport this fixup to `msys2-3_2_0-release` first.

lazka: this is in msys2-3_2_0-release and this PR now from what I see

dscho: Sorry, I forgot to update this thread ;-)

PR #64, opened 2021-11-08
kill: avoid using a potentially uninitialized value

dscho:
This clear bug was pointed out by @jeremyd2019 in the PR to upgrade to Cygwin v3.3.2: https://github.com/msys2/msys2-runtime/pull/63/files/950533d95e26b54e2e1ab26d4f5e65ce0c3f7305#r745004385


dscho:
My intention is not so much to take this to a new `msys2-runtime` version based on v3.2.0, but to forward-port it to #63 and then continue with working on `msys2-runtime` v3.3.2.

jeremyd2019 (approved):

PR #69, opened 2021-11-18
More backports of Cygwin fixes

dscho:
From my brief reading, the `STATUS_PENDING` one seems to be relatively important.

An alternative to taking this PR would be to wait for Cygwin v3.3.3. I have no information how imminent that version is, though.


lazka:
Closing in favor of #71

PR #72, opened 2021-12-10
 Route "Device Status Report" responses correctly

dscho:
As reported in https://github.com/git-for-windows/git/issues/3579, when using MSYS2 runtime v3.3.3 in a regular Git Bash, after `git config -e --global` launched `vim` and the editor was closed, seemingly random characters are "ghost-typed". Concretely, I saw `1R77;30502;0c10;rgb:bfbf/bfbf/bfbf11;rgb:0000/0000/0000` being "ghost-typed" every time Git had called vi for editing a file.

To be precise, the hexdump -C output is this:

```
00000000  1b 5b 32 3b 32 52 1b 5b  33 3b 31 52 1b 5b 3e 37  |.[2;2R.[3;1R.[>7|
00000010  37 3b 33 30 35 30 32 3b  30 63 1b 5d 31 30 3b 72  |7;30502;0c.]10;r|
00000020  67 62 3a 62 66 62 66 2f  62 66 62 66 2f 62 66 62  |gb:bfbf/bfbf/bfb|
00000030  66 07 1b 5d 31 31 3b 72  67 62 3a 30 30 30 30 2f  |f..]11;rgb:0000/|
00000040  30 30 30 30 2f 30 30 30  30 07                    |0000/0000.|
```

This is the response for a `CSI [ 6n` request, and the response was clearly routed to the terminal by mistake instead of the process that requested it.

The regression was actually introduced into v3.2.0, but not noticed earlier. [I reported the issue](https://sourceware.org/pipermail/cygwin-patches/2021q4/011587.html) after trying to figure out a workaround for three days, without success, and the original author of the commit introducing the regression provided a patch that works around it. I do not yet understand the patch, but let's take the patch for now because the regression is pretty severe.

While at it, also include a fixup that handles two previously unhandled instances of the `cygwin-pty` pipe (which is renamed to `msys-pty` in MSYS2) that I spotted while working on this bug.

### DIFF
--- a/winsup/cygwin/fhandler_tty.cc
+++ b/winsup/cygwin/fhandler_tty.cc
@@ -1153,7 +1153,11 @@ fhandler_pty_slave::reset_switch_to_pcon (void)
 		    {
 		      char pipe[MAX_PATH];
 		      __small_sprintf (pipe,
+#ifdef __MSYS__
+			       "\\\\.\\pipe\\msys-%S-pty%d-master-ctl",
+#else
 			       "\\\\.\\pipe\\cygwin-%S-pty%d-master-ctl",
+#endif
 			       &cygheap->installation_key, get_minor ());
 		      pipe_request req = { GetCurrentProcessId () };
 		      pipe_reply repl;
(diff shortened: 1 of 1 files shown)


lazka:
-> https://github.com/msys2/MSYS2-packages/pull/2730

PR #73, opened 2021-12-12
mapped drive symlink fix

### LINKED ISSUE #58: Can't access network drives (shared folders)
Hi, I apologize if this is the wrong place to post this issue.

I recently installed MSYS2 on my Windows 10 VM in VirtualBox, and I quickly found out that I can't navigate to any of my shared folders (which are mounted as Network Drives in Windows). The primary error message is `Too many levels of symbolic links`.

Here is some raw output from the shell:
```
$ cd /t
bash: cd: /t: Too many levels of symbolic links

$ df
df: /b: Too many levels of symbolic links
df: /g: Too many levels of symbolic links
df: /j: Too many levels of symbolic links
df: /p: Too many levels of symbolic links
df: /t: Too many levels of symbolic links
Filesystem        1K-blocks     Used Available Use% Mounted on
C:/Software/msys2  51849068 29908016  21941052  58% /
D:                    59642    59642         0 100% /d
```

Drive D is a CD-drive and I am able to access it normally, as well as the C drive (`/c`). All of my projects are on the host system so I need to access them via mechanisms such as shared folders, which are implemented as network drive mounts in VirtualBox's Windows guests.

I am able to use `cmd` and `powershell` normally in network drives as well as many other programs, this is one of the rare occasions where they have caused me trouble, I don't even recall the last issue I had with them.

jeremyd2019:
This is supposed to fix #58.

I haven't had a chance to test it yet, but putting it out here in case somebody else gets a chance first

### DIFF
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -3670,29 +3670,43 @@ restart:
 		  if (RtlEqualUnicodePathPrefix (&fpath, &ro_u_uncp, TRUE)
 		      && !RtlEqualUnicodePathPrefix (&upath, &ro_u_uncp, TRUE))
 		    {
-		      /* ...get the remote path from the volume path name,
-			 replace remote path with drive letter, check again. */
+		      /* ...get the remote path, replace remote path
+			 with drive letter, check again. */
+		      WCHAR drive[3] =
+			{(WCHAR) towupper (upath.Buffer[4]), L':', L'\0'};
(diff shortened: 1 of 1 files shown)


jeremyd2019:
I never did manage to test this myself, but a user has confirmed that it works.  https://github.com/msys2/MSYS2-packages/issues/2753#issuecomment-998365867

PR #79, opened 2022-01-09
Special case the environment variables SHELL and ORIGINAL_PATH

dscho:
In MSYS2 <-> MINGW transitions, we convert a couple of environment variables between Unix and Windows styles, most notably `PATH`.

In the Git for Windows project, we realized that the environment variables `SHELL` and `ORIGINAL_PATH` need the same treatment. These patches have proved their worth for years, and it is time for them to be integrated back into the actual MSYS2 runtime.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,7 +15,7 @@ jobs:
         with:
           msystem: MSYS
           update: true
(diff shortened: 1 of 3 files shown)


dscho:
@lazka this PR is ready now, sorry for missing the CI failure earlier.

dscho:
@lazka any objections against merging?

lazka:
lgtm

PR #82, opened 2022-01-17
path_conv: a directory with `.lnk` suffix is not a `.lnk` file

### LINKED ISSUE #81: Directories ending with ".lnk" should not be handled as links
I want to create two directories, "foo" and "foo.lnk".

The following command sequence works fine:
```
alb@bla MINGW64 /d/TEMP
$ mkdir foo

alb@bla MINGW64 /d/TEMP
$ mkdir foo.lnk

alb@bla MINGW64 /d/TEMP
$ ls
foo/  foo.lnk/
```

But if the directory ending with ".lnk" is created first, we run into an error:
```
alb@bla MINGW64 /d/TEMP
$ mkdir foo.lnk

alb@bla MINGW64 /d/TEMP
$ mkdir foo
mkdir: cannot create directory ‘foo’: File exists
```

Presumably, ``foo.lnk`` is handled as link (or shortcut) due to the .lnk suffix.

Windows shortcuts (.lnk files) are always *files*, and directories ending with .lnk are just directories.
Msys2 should not handle such directories as links.

While the above example is a constructed one, here's the issue which I stumbled upon: It's not possible to clone the following git repository with msys2 git: 

```
alb@bla MSYS /d/TEMP
$ git clone --depth 1 git://code.qt.io/qt/qtbase.git
...output omitted, but this will fail. See the next command why that is.

alb@bla MSYS /d/TEMP/qtbase
$ git reset --hard
fatal: cannot create directory at 'tests/auto/corelib/io/qdir/testdir/dir.lnk/subdir': File exists
```

Git failed to create a certain directory. Because the directory ``tests/auto/corelib/io/qdir/testdir/dir.lnk/subdir.lnk`` is also part of the repo and was created before.

Side note: Interestingly, git from https://gitforwindows.org/ can clone this repo just fine, though it's also msys2-based.

dscho:
This clears a variable when we half-detected a symlink emulated via a `.lnk` file (and when it turned out that it was a directory instead).

This fixes https://github.com/msys2/msys2-runtime/issues/81

### DIFF
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -3498,6 +3498,7 @@ restart:
 	 hasn't been found. */
       if (ext_tacked_on && !had_ext && (fileattr & FILE_ATTRIBUTE_DIRECTORY))
 	{
+	  fileattr = INVALID_FILE_ATTRIBUTES;
 	  set_error (ENOENT);
 	  continue;
 	}


dscho:
As per https://github.com/msys2/msys2-runtime/issues/81#issuecomment-1015245013, this does what it is supposed to do.

Issue #83, opened 2022-01-23
Ctrl-C interrupts the whole ssh session while catching by in interactive mode


pnuzhdin:
The issue:
1. Open `cmd.exe`.
2. Run `C:\msys64\usr\bin\ssh.exe -v somehost`.
3. While in the interactive session run some long-running process e.g. `top`.
4. Press `Ctrl+C` to interrupt the long-running process inside the interactive session.

Observe: The whole interactive ssh session is terminated with the following debug output:
```
debug1: channel 0: free: client-session, nchannels 1
Killed by signal 2.
```

At the same time if you run another MSYS2 package e.g. `vim` instead of `ssh`, `Ctrl+C` works correctly so I'm not sure does it relate to MSYS2 runtime or just OpenSSH package.

crrodriguez:
Observing the same behaviour with ssh.exe shipped with git.. this breaks all workflow for me.

dscho:
Can you try again after setting `MSYS=enable_pcon`?

pnuzhdin:
> Can you try again after setting `MSYS=enable_pcon`?

Thanks, it fixes the issue for me!

crrodriguez:
> Can you try again after setting `MSYS=enable_pcon`?

that restores the correct behaviour, thanks.

jeremyd2019:
Just to clarify, are you talking about running ssh from cmd directly, or running cmd from bash, and then running ssh in that cmd?  I tried to reproduce with the former and it didn't seem to reproduce.  But there was a commit to upstream cygwin recently that claims to fix a cygwin -> native -> cygwin Ctrl-C issue that seems like it might be related

Update: even from bash -> cmd -> msys2 ssh I still can't dupe it

jeremyd2019:
I can dupe with Git for Windows but not with "stock" msys2.

dscho:
@pnuzhdin please test with [the artifact](https://github.com/msys2/MSYS2-packages/suites/5310719289/artifacts/165102515) from the PR with the proposed fix that is under review.

dscho:
Well, now you can simply update...

pnuzhdin:
@dscho @jeremyd2019 thanks! I've updated msys and now it works properly even without `MSYS=enable_pcon`

pnuzhdin:
It seems this issue is reproduced again for:
```
msys2-launcher 1.5-1
msys2-runtime 3.3.5-1
```

And having `MSYS=enable_pcon` doesn't help.

[...]

PR #86, opened 2022-02-14
backport recent console patches

jeremyd2019:
This seems to solve the same issue that git-for-windows/msys2-runtime@1d0701d attempted to solve, but without the side effects (git-for-windows/git#3657, #83).

### DIFF
--- a/winsup/cygwin/fhandler.h
+++ b/winsup/cygwin/fhandler.h
@@ -1941,6 +1941,8 @@ class fhandler_termios: public fhandler_base
     fh->copy_from (this);
     return fh;
   }
+  static bool path_iscygexec_a (LPCSTR n, LPSTR c);
+  static bool path_iscygexec_w (LPCWSTR n, LPWSTR c);
 };
 
 enum ansi_intensity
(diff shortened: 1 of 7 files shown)


dscho:
Thank you! I integrated these patches via https://github.com/git-for-windows/msys2-runtime/pull/40#commits-pushed-c8568e0, let's hope that the Console problems settle down some (my expectations for that are low, though).

dscho:
I can confirm that this indeed fixes https://github.com/GitCredentialManager/git-credential-manager/issues/576, too.

dscho (approved):

jeremyd2019:
I backported the commits to git-for-windows/msys2-runtime main and confirmed that it fixes #83 and git-for-windows/git#3657 there (though your PR updating to 3.3.4 makes more sense, because there were merge issues backporting to 3.3.3 resulting in my dropping some fixes that depended on other fixes that weren't in 3.3.3).

jeremyd2019:
One thing to look out for - I think the proliferation of GenerateConsoleCtrlEvent calls in these patches and in some already in 3.3.4 may degrade the custom Ctrl-C handling code that was added to msys2-runtime.

PR #95, opened 2022-05-27
Convert the trailing colon to semi-colon while converting paths

### LINKED ISSUE #94: msys2_path_conv: Don't strip trailing `:`
While debugging https://github.com/msys2/MSYS2-packages/issues/3014#issuecomment-1135921380 I found that trailing `:` has a different meaning and that the runtime just strips those (which is the root cause of the issue). I would expect the runtime to convert `:` to `;` and leave it there at the end.

```
❯ TESTENV=/tmp/hello/world:/tmp/new: python -c 'import os; print(os.environ["TESTENV"])'
C:\msys64\tmp\hello\world;C:\msys64\tmp\new
```
I would expect the output to be `C:\msys64\tmp\hello\world;C:\msys64\tmp\new;`.  Is it something that people would expect or can be treated as an expected behaviour? Does this break any use case?

naveen521kk:
Previously, trailing colons were stripped which caused some issues,
with texinfo package, where trailing `:` or `;` would mean adding standard
library paths, and stripping it causes errors. Logically, converting those
trailing colons to semi-colons seems right rather than stripping them.

Fixes https://github.com/msys2/msys2-runtime/issues/94
Also see, https://github.com/msys2/path_convert/pull/5

### DIFF
--- a/winsup/cygwin/msys2_path_conv.cc
+++ b/winsup/cygwin/msys2_path_conv.cc
@@ -602,8 +602,6 @@ void ppl_convert(const char** from, const char* to, char** dst, const char* dste
 
     if (!prev_was_simc) {
         subp_convert(&beg, it, is_url, dst, dstend);
-    } else {
-        *dst -= 1;
     }
 }
 


naveen521kk:
Locally tested with the built from CI and works as expected.
```
$ ABC=:/hello:/usr/lib/:/var: py -c 'import os; print(os.environ["ABC"])'
;C:\dev\texlive-2022\msys64\hello;C:\dev\texlive-2022\msys64\usr\lib\;C:\dev\texlive-2022\msys64\var;
```
Not sure whether it'll break anyone else's workflow though, since this code isn't touched for a long time now.

dscho (approved):
Apart from the typo "striping" (should be "stripping" instead) looks good!

naveen521kk:
> Apart from the typo "striping" (should be "stripping" instead) looks good!

Thanks for the review! I've fixed the typo. 

I think this should be squashed into https://github.com/msys2/msys2-runtime/commit/03929e0952ba0d460b0348046c9354fcacce0b78 on the next rebase.

dscho:
> I think this should be squashed into [03929e0](https://github.com/msys2/msys2-runtime/commit/03929e0952ba0d460b0348046c9354fcacce0b78) on the next rebase.

I agree. Would you mind marking it up as a `fixup!` commit so that `git rebase --autosquash` will pick it up correctly?

naveen521kk:
> Would you mind marking it up as a `fixup!` commit so that `git rebase --autosquash` will pick it up correctly?

I'll do it in a few hours.

dscho (approved):

dscho:
Thank you!

dscho:
I opened https://github.com/msys2/MSYS2-packages/pull/3021 with the corresponding patch.

naveen521kk:
Thanks for reviewing!

PR #97, opened 2022-07-03
uname: allow setting the system name to CYGWIN

lazka:
We are currently trying to move our cygwin build environment closer
to cygwin and some autotools/bash based build systems call "uname -s"
to figure out the OS and in many cases only handle the cygwin case, so
we have to patch them.

With this instead of patching we can set MSYSTEM=CYGWIN and change
uname output that way.

The next step would be to always output CYGWIN in an msys env by default,
but for now this allows us to get rid of all the patches without
affecting users.

### DIFF
--- a/winsup/cygwin/uname.cc
+++ b/winsup/cygwin/uname.cc
@@ -23,6 +23,24 @@ extern "C" int getdomainname (char *__name, size_t __len);
 #define ATTRIBUTE_NONSTRING
 #endif
 
+static const char*
+get_sysname()
+{
+#ifdef __MSYS__
+  char* msystem = getenv("MSYSTEM");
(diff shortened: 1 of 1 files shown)


lazka:
There is also `uname -o` -> `Msys`, which I'm not sure about... I'll have another look there.

lazka:
> There is also `uname -o` -> `Msys`, which I'm not sure about... I'll have another look there.

ah, that's hardcoded in coreutils, so nothing we can do here.

lazka:
The "uname -o" part is handled here now: https://github.com/msys2/MSYS2-packages/pull/3080

dscho (approved):
Looking good!

lazka:
thanks!

Issue #99, opened 2022-09-05
Empty variables are silently deleted.


eli-schwartz:
Windows is perfectly okay with environment variables of the form `"MESON_SUBDIR="`, at least I can set such a variable in e.g. a python script and communicate it to another process. Although cmd.exe's `set` IIUC will delete it? But msys2 is not a cmd.exe emulator...

However, trying to run this fails:
```
"env" "MY_ENV=1" "MESON_SOURCE_ROOT=D:\a\meson\meson\test cases\common\51 run target" \
    "MESON_BUILD_ROOT=D:\a\meson\meson\tmp59june4u" "MESON_SUBDIR=" \
    "MESONINTROSPECT=D:\a\_temp\msys64\mingw32\bin\python3.exe D:\a\meson\meson\meson.py introspect" \
    "D:\a\_temp\msys64\mingw32\bin\python3.exe" "D:/a/meson/meson/test cases/common/51 run target/check-env.py" \
    "../test cases/common/51 run target" "." "../test cases/common/51 run target/"
```

```
Traceback (most recent call last):
  File "D:\a\meson\meson\test cases\common\51 run target\check-env.py", line 8, in <module>
    assert 'MESON_SUBDIR' in os.environ
AssertionError
```

This was brought up e.g. in https://github.com/msys2/MINGW-packages/issues/1599 in the past.

...

It seems very wrong for the msys2 runtime to do this. If I run `python3.exe basically-env.py "MESON_SUBDIR=" <script>` it works flawlessly, but if I run `env.exe "MESON_SUBDIR=" <script>` it refuses to work because it's trying to emulate cmd.exe instead of Unix?

Unfortunately, Meson *needs* this functionality, because "the concept of being empty" is crucial information that needs to be passed from one process to another. I would like to pass this information via the most efficient program for setting such information, but it looks like I might need to restrict that to "real unixes".

Thoughts? Is it possible this restriction can be lifted?

dscho:
I _think_ the original rationale was that setting an environment variable to an empty value deletes it in `CMD`, and that this behavior should be replicated by the MSYS2 runtime.

Over in Git for Windows, we do rely on the ability to be able to pass empty values for environment variables, that's there is https://github.com/git-for-windows/msys2-runtime/commit/3e9c003101b17d11e250c8a3ec92de219db67860 in Git for Windows' fork of the MSYS2 runtime (and we've had it since 2015, the commit date of _that_ commit notwithstanding).

[...]

PR #101, opened 2022-09-06
Allow empty-valued environment variables to be passed to non-MSYS programs

dscho:
This addresses https://github.com/msys2/msys2-runtime/issues/99 and also reduces the difference between the MSYS2 runtime and Git for Windows' fork of the same.

### DIFF
--- a/winsup/cygwin/environ.cc
+++ b/winsup/cygwin/environ.cc
@@ -36,6 +36,7 @@ static char **lastenviron;
 /* Parse CYGWIN options */
 
 static NO_COPY bool export_settings = false;
+static bool emptyenvvalues = true;
 
 enum settings
   {
@@ -130,6 +131,7 @@ static struct parse_thing
   {"enable_pcon", {&disable_pcon}, setnegbool, NULL, {{true}, {false}}},
   {"winjitdebug", {&winjitdebug}, setbool, NULL, {{false}, {true}}},
   {"nativeinnerlinks", {&nativeinnerlinks}, setbool, NULL, {{false}, {true}}},
+  {"emptyenvvalues", {&emptyenvvalues}, setbool, NULL, {{false}, {true}}},
   {NULL, {0}, setdword, 0, {{0}, {0}}}
 };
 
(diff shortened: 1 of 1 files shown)


lazka (approved):
lgtm

just wondering, do you think if this turns out to not make any problems (after some time) could we propose it to cygwin without the escape hatch?

dscho:
> do you think if this turns out to not make any problems (after some time) could we propose it to cygwin without the escape hatch?

Why without the escape hatch?

But yes, my answer is: yes, I think we should upstream this.

lazka:
> Why without the escape hatch?

if no one knows why it was added and if there is no fallout in MSYS2 then it seems like unnecessary complexity. we'll see....

dscho:
> > Why without the escape hatch?
> 
> if no one knows why it was added and if there is no fallout in MSYS2 then it seems like unnecessary complexity. we'll see....

Well, the original behavior (to treat environment variables with empty values as if they were deleted) hails from Cygwin, not from MSYS2... So they would know (or should know) why it was added.

lazka:
There is some fallout from this, some fixable, some I don't understand yet.

We currently have CPython in MSYS2 mode if "MSYSTEM" isn't set and in various examples we just do "MSYSTEM= python" because up until now this just happened to remove the env var I guess.

That would be fixable by users by changing the code to `unset MSYSTEM`, but for some reason that doesn't work (likely didn't work before as well?):

```console
$ unset MSYSTEM
$ env | grep MSYSTEM=
MSYSTEM=MINGW64
```

So, the Python thing we can likely fix in MSYS2 by also recognizing an empty MSYSTEM env var in CPython -> https://github.com/msys2-contrib/cpython-mingw/pull/107

Any ideas why unset doesn't work here?

lazka:
* Python is fixed here: https://github.com/msys2/MINGW-packages/pull/13251
* The unset issue is tracked here: https://github.com/msys2/MSYS2-packages/issues/3200

I'll open issues for anything else that comes up.

PR #103, opened 2022-09-20
path_conv: be a lot more careful when to keep a trailing slash

dscho:
On FAT file systems, the normalized path is used to fake inode numbers. As a result, `Y:\directory\` and `Y:\directory` have different inode numbers!!!

One consequence of this bug has been reported on the Git for Windows: the `find` and `rm` commands can error out on FAT file systems with very confusing "No such file or directory" errors, for no good reason (see https://github.com/git-for-windows/git/issues/1497 and https://github.com/git-for-windows/git/issues/1536).

This bug had been fixed in Cygwin as early as 1997... and we reintroduced it in MSYS2.

Let's fix it again.

This patch has been carried in Git for Windows for several years, with no further bug reports.

While at it, replace the very sad, sad, grammatically-incorrect one-liner of a commit message with a detailed explanation of the context, the intent and the justification of the code change.

### DIFF
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1254,17 +1254,17 @@ path_conv::check (const char *src, unsigned opt,
 		cfree (wide_path);
 	      wide_path = NULL;
 	    }
-	}
 
-      if (need_directory)
-	{
-	  size_t n = strlen (this->path);
-	  /* Do not add trailing \ to UNC device names like \\.\a: */
-	  if (this->path[n - 1] != '\\' &&
-	      (strncmp (this->path, "\\\\.\\", 4) != 0))
+	  if (need_directory)
 	    {
-	      this->modifiable_path ()[n] = '\\';
-	      this->modifiable_path ()[n + 1] = '\0';
+	      size_t n = strlen (this->path);
+	      /* Do not add trailing \ to UNC device names like \\.\a: */
+	      if (this->path[n - 1] != '\\' &&
+		  (strncmp (this->path, "\\\\.\\", 4) != 0))
+		{
+		  this->modifiable_path ()[n] = '\\';
+		  this->modifiable_path ()[n + 1] = '\0';
+		}
 	    }
 	}
 


mati865:
> While at it, replace the very sad, sad, grammatically-incorrect one-liner of a commit message with a detailed explanation of the context, the intent and the justification of the code change.

Seems the commit with this change is missing.

dscho:
> > While at it, replace the very sad, sad, grammatically-incorrect one-liner of a commit message with a detailed explanation of the context, the intent and the justification of the code change.
> 
> Seems the commit with this change is missing.

@mati865 This is not a separate commit, and it is not missing, it is https://github.com/msys2/msys2-runtime/pull/103/commits/bb6e68e35cdc71aacc06badc7e58fe053e7f366a.

The special form `amend!` is generated by [`git commit --fixup reword:<commit>`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---fixupamendrewordltcommitgt) (which usually does not allow committing any actual code changes, but you can `git commit --amend` it and it works as intended, which is what I did here).

mati865:
Oh sorry, I was under impression you intended to change in-code comment.

dscho:
> Oh sorry, I was under impression you intended to change in-code comment.

@mati865 are you referring to this comment?

```c
/* Do not add trailing \ to UNC device names like \\.\a: */
```

If so, this is actually a comment in Cygwin (i.e. not specific to MSYS2, and hence outside the scope of this here PR): https://github.com/cygwin/cygwin/blob/master/winsup/cygwin/path.cc#L1253

mati865:
Yeah, I have misunderstood the intention.

dscho:
@lazka could you have a look, please?

lazka:
Because I was confused too -> this amends 4676ec44f24a0

I'm afraid I don't quite understand why the restoring of the trailing slash has anything to do with PC_NOFULL. (and what does this function itself even do?.. and many more questions..)

But since it reduces the diff to cygwin while the originally intended behavior change is still there this looks good to me.

lazka:
In light of #115, which is partly triggered by 4676ec4, do we know what 4676ec4 was trying to fix? i.e. why is `/ -> C:/msys64/` good and `/ -> C:/msys64` bad? I'm not trying to block this with my question though, just wondering..

dscho:
> Because I was confused too -> this amends [4676ec4](https://github.com/msys2/msys2-runtime/commit/4676ec44f24a0bfa208e6f0c79b0b059c88f6b1a)

Good point, sorry for not linking to this myself. It is of course part of the context required to understand this change.

> I'm afraid I don't quite understand why the restoring of the trailing slash has anything to do with PC_NOFULL. (and what does this function itself even do?.. and many more questions..)

[`PC_NOFULL` tells `path_conv` to keep relative paths relative](https://github.com/msys2/msys2-runtime/blob/f975120d20d51ca11217e4abbf16cc51ab8c9290/winsup/cygwin/path.h#L52).

So a valid question is: why is this `need_directory` flag only heeded in that mode?

I'll dig deeper into this question in a moment, but let me first note that [this is upstream Cygwin's current behavior](https://github.com/msys2/msys2-runtime/blob/cygwin-3_3_6-release/winsup/cygwin/path.cc#L1239-L1261): the `need_directory` flag [is set if the user looked for a directory](https://github.com/msys2/msys2-runtime/blob/cygwin-3_3_6-release/winsup/cygwin/path.cc#L703-L710), is [then used to append a slash if the file attributes could not be determined](https://github.com/msys2/msys2-runtime/blob/cygwin-3_3_6-release/winsup/cygwin/path.cc#L1178-L1190) and eventually is used _only_ in the case we want to convert a relative path to a relative path.

Note that this `PC_NOFULL` mode is triggered by `CCP_RELATIVE` which is used almost everywhere anyway.

Now, let me dig a bit deeper where that logic comes from. The monster patch 3f3bd10104550243781f0b4d9248975e35d91ac7 touches that part, but really only changes indentation (check for yourself via `git show --patience -w 3f3bd10104550243781f0b4d9248975e35d91ac7 -- winsup/cygwin/path.cc`), likewise 1870c688063bc20ae29a9b2744471705d790fe47 only reformats said code, same for 81adfe28623b8a40d8b946c7e9355e3e78847cf8. The parent commit [still has the same logic: the slash is only appended if `opt & PC_NOFULL` and `need_directory` are true](https://github.com/msys2/msys2-runtime/blob/81adfe28623b8a40d8b946c7e9355e3e78847cf8%5E/winsup/cygwin/path.cc#L1246-L1261). Re-starting the analysis from there (via `git log -w -L 1246,1261:winsup/cygwin/path.cc 81adfe28623b8a40d8b946c7e9355e3e78847cf8^`), we are led to believe that d3c7e9de2823e072732fb9598502c4ebd2a1d0fb introduced that logic, but that's not so: it was moved from [here](https://github.com/msys2/msys2-runtime/commit/d3c7e9de2823e072732fb9598502c4ebd2a1d0fb#diff-92ef3f1ca262cde946d683edadd29900639e38351757d25476c2c0b2a9a7debbL1217-L1231) to [here](https://github.com/msys2/msys2-runtime/commit/d3c7e9de2823e072732fb9598502c4ebd2a1d0fb#diff-92ef3f1ca262cde946d683edadd29900639e38351757d25476c2c0b2a9a7debbR648-R660). And if we try to dig deeper from there, we come to a hard stop at 1fd5e000ace55b323124c7e556a7a864b972a5c4, which is super unhelpful because it is a wholesale tar import from the year 2000 when [we have been told that the bug fix happened in 1997](https://github.com/git-for-windows/git/issues/1497#issuecomment-372665870).

So I guess that the best thing we can do is to trust Cygwin which has kept this `need_directory` handling guarded by the `PC_NOFULL` flag, and we should not deviate from it, _in particular_ because we know that this deviation [caused a bug](https://github.com/git-for-windows/git/issues/1497) where the inode calculation (necessary on non-NTFS drives) would be inconsistent causing `find` to be surprised to not find a file that it had _just_ found.

> But since it reduces the diff to cygwin while the originally intended behavior change is still there this looks good to me.

Indeed. That was my primary goal here.

> In light of #115, which is partly triggered by [4676ec4](https://github.com/msys2/msys2-runtime/commit/4676ec44f24a0bfa208e6f0c79b0b059c88f6b1a), do we know what [4676ec4](https://github.com/msys2/msys2-runtime/commit/4676ec44f24a0bfa208e6f0c79b0b059c88f6b1a) was trying to fix? i.e. why is `/ -> C:/msys64/` good and `/ -> C:/msys64` bad? I'm not trying to block this with my question though, just wondering..

The earliest reference I can find is msys2/MSYS2-packages@c560737d4208d58bdc802cb5c63713fda516ce5e. It is painful to think of the reason why we cannot ask the author of that patch for more information, and the commit message does not really provide said information, either.

Having said that, this commit talks about the goal (without providing any motivating context) _but does not move the check out of the `PC_NOFULL` clause_. That move was done in [msys2/MSYS2-packages@175489d513f62aa456d9d4699692d1d734a6ba74](https://github.com/msys2/MSYS2-packages/commit/175489d513f62aa456d9d4699692d1d734a6ba74#diff-51f6426dabc23e1f73bd32af50bdb069da2d099b20215955d7c192c8239a1c33). And in that patch message we finally [get the reasoning for moving the logic out of the `PC_NOFULL` clause](https://github.com/msys2/MSYS2-packages/commit/175489d513f62aa456d9d4699692d1d734a6ba74#diff-51f6426dabc23e1f73bd32af50bdb069da2d099b20215955d7c192c8239a1c33R11-R16):

> This fixes path conversion of the root mount, e.g.: '/' -> 'C:/msys64' .. in this example, the result of converting '/' would be
'C:\msys64' when it clearly should be 'C:\msys64\', I say clearly because it is obvious that converting a path that ends in / should result in a path ending in \ every time.

On the face of it, this is plausible. But it becomes a lot less plausible when you keep in mind that this path conversion is also used for the inode emulation that calculates a hash from the absolute path and therefore _requires_ the conversion to be consistent whether the original path has the trailing slash or not. We do not want `/directory` to have "a different inode" than `/directory/`, after all.

dscho:
@lazka what do you think? Run with this, reducing the diff relative to Cygwin?

lazka (approved):

lazka:
Thanks for the thorough investigation (I tried to do something similar but gave up).

I'm fine with either applying this, or removing the original patch completely.

dscho:
> I'm fine with either applying this, or removing the original patch completely.

I am reluctant to remove the original patch entirely because I suspect that there _was_ some merit in doing it in the first place.

lazka:
I guess I found the reason why this was added in the first, place. The GCC build depends on slashes surviving: https://github.com/msys2/MINGW-packages/pull/14163 Maybe it can be fixed there.

lazka:
> I guess I found the reason why this was added in the first, place. The GCC build depends on slashes surviving: [msys2/MINGW-packages#14163](https://github.com/msys2/MINGW-packages/pull/14163) Maybe it can be fixed there.

A less invasive "fix": #117

PR #110, opened 2022-10-21
Fix `Bad address` when running `cmd /c [...]`

### LINKED ISSUE #108: Regression introduced in msys2-runtime-3.3.6-1 when executing cmd /c
There has been a regression in msys2-runtime when it comes to invoking `cmd.exe`

Here is the old expected behaviour with msys2-runtime-3.3.5-3 or earlier:
```
pete@good_msys MINGW32 ~
$ uname -a
MINGW32_NT-10.0-22621 good_msys 3.3.5-341.x86_64 2022-07-04 21:33 UTC x86_64 Msys

pete@good_msys MINGW32 ~
$ cmd.exe /c 'echo BLAH'
BLAH
```

And this is what happens when upgrading to msys2-runtime-3.3.6-1 or later:
```
pete@bad_msys MINGW32 ~
$ uname -a
MINGW32_NT-10.0-22621 bad_msys 3.3.6-341.x86_64 2022-09-20 22:07 UTC x86_64 Msys

pete@bad_msys MINGW32 ~
$ cmd.exe /c 'echo BLAH'
bash: /c/Windows/System32/cmd.exe: Bad address
```

This issue has been confirmed by other people (on the Discord channel)  and whereas, for the simple example above, one workaround has been found, which consists of doubling the '/' (`cmd.exe //c 'echo BLAH'`), such a workaround does not appear to be effective with more complex command invocations, [such as one I am using in a bash script where `signtool.exe` must be invoked in order to digitally sing a PowerShell script](https://github.com/pbatard/Fido/blob/master/sign.sh#L24). In this case, when using `//`,  the `Bad address` error  becomes `'\"C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x64\signtool\"' is not recognized as an internal or external command, operable program or batch file`.

I am therefore opening this issue to ask the msys2-runtime developers to investigate the changes that occurred between 3.3.5-3 and 3.3.6-1, in order to fix this regression, so that invoking a complex command such as `cmd.exe /c '"C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x64\signtool" sign /v /sha1 3dbc3a2a0e9ce8803b422cfdbc60acd33164965d /fd SHA256 /tr http://sha256timestamp.ws.symantec.com/sha256/timestamp /td SHA256 Fido.ps1'` from bash will work as it did before.

dscho:
In 2b4f986e49 (Cygwin: pty: Treat *.bat and *.cmd as a non-cygwin console app., 2022-07-31), we introduced a bug fix that  specifically looks for a suffix of the command's file name.

However, that file name might be set to `NULL`, namely when `null_app_name == true`, which is the case when we detected a command-line `cmd /c [...]`.

But the commit mentioned above did not account for that possibility, instead assuming that it always has to check the file name for a `.bat` or `.cmd` suffix. As a consequence, `cmd /c [...]` invocations are completely broken in v3.3.6, resulting in a `Bad address` error.

Let's guard the code properly so that it does not try to look at the file name suffix of `(WCHAR *)NULL`.

This fixes https://github.com/msys2/msys2-runtime/issues/108

### DIFF
--- a/winsup/cygwin/spawn.cc
+++ b/winsup/cygwin/spawn.cc
@@ -198,6 +198,8 @@ handle (int fd, bool writing)
 static bool
 is_console_app (WCHAR *filename)
 {
+  if (filename == NULL)
+    return true; /* The command executed is command.com or cmd.exe. */
   HANDLE h;
   const int id_offset = 92;
   h = CreateFileW (filename, GENERIC_READ, FILE_SHARE_READ,


dscho:
I [also sent this patch to Cygwin](https://inbox.sourceware.org/cygwin-patches/8rqs6n82-0oq9-2200-944n-74s7o699385o@tzk.qr/).

jeremyd2019:
Thanks @dscho! I was planning to look into this, but probably wouldn't have had time until Monday

lazka:
It was fixed slightly differently in cygwin. Which version should we use?

https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=commitdiff;h=08281cf4cca9593adcc3d30184322dc60fa1cd61

mati865:
They are equal in functionality.

dscho:
> It was fixed slightly differently in cygwin. Which version should we use?
> 
> https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=commitdiff;h=08281cf4cca9593adcc3d30184322dc60fa1cd61

Yes, it was a _highly_ frustrating experience. @tyan0, instead of reviewing my patch and working with me so that I can fix whatever issues they find with the patch, simply replaced it, without anything in the way of communication clearly and openly (there was just a note that it was replaced, that was all).

Sadly the replacement patch [was pushed to the `cygwin-3_3-branch` already](https://github.com/cygwin/cygwin/commits/cygwin-3_3-branch), without giving anybody any chance to improve it. And there is a ton of room for improvement:
the commit message is seriously wanting, not even using the information I provided in my proposal, let alone offering any adequate explanation as to what the "issue" is that it talks about in a sentence whose grammar makes it hard to parse:

```
Cygwin: pty: Fix 'Bad address' error when running 'cmd.exe /c dir'

- If the command executed is 'cmd.exe /c [...]', runpath in spawn.cc
  will be NULL. In this case, is_console_app(runpath) check causes
  access violation. This case also the command executed is obviously
  console app., therefore, treat it as console app to fix this issue.
```

Since that commit is already set in stone by virtue of being pushed to the `cygwin-3_3-branch`, we are stuck with it.

dscho:
On Sat, 22 Oct 2022, Mateusz Mikuła wrote:

> They are equal in functionality.

No, they are not. In my version, we still returned `false` in case of `command.com` and `cmd.exe` being called with `/c <command>`. In @tyan0's version, it returns `true`.

mati865:
Yeah, I had noticed it too late so deleted comment to avoid confusing others.

dscho:
Just to add insult to injury, it [looks as if the counter-proposal that was pushed directly to `cygwin-3_3-branch` is incorrect, too](https://inbox.sourceware.org/cygwin-patches/Y1ZazH6objN99mSz@calimero.vinschen.de/).

PR #116, opened 2022-11-08
ci: avoid using Node.js 12 Actions

dscho:
I noticed [in this build](https://github.com/msys2/msys2-runtime/actions/runs/3420824668#annotations) that we're using outdated Actions. Let's upgrade them and avoid having our CI fail due to use of deprecated Node.js versions

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,7 +8,7 @@ jobs:
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


lazka (approved):

dscho:
I won't accompany this PR with one over at msys2/MSYS2-packages because this change is irrelevant to building the `msys2-runtime` package.

PR #118, opened 2022-11-23
fix(error message): Clarify magic number mismatch error message between heap child information

Quenty:
msys2 is a fork of cygwin but the error message for magic_versions is cygwin specific. Removing this message may remove a lot of documentation for users searching online.

However, the documentation itself is not useful. Users will not find a cygwin version nor does the help online specify exact information. While cygwin was intended to be installed at a single location, msys2 is used in a variety of stand alone tools, one of which is git-for-windows. Whenever a version mismatch can occur this can lead to programs such as VS Code breaking other git installations, or vice-versa. In this regard a bit of a better error message can help people figure out what is wrong on their msys2 or cygwin installed codebases.

Docs PR:
https://github.com/msys2/msys2.github.io/pull/235

### DIFF
--- a/winsup/cygwin/dcrt0.cc
+++ b/winsup/cygwin/dcrt0.cc
@@ -1360,12 +1360,25 @@ multiple_cygwin_problem (const char *what, uintptr_t magic_version, uintptr_t ve
     system_printf ("%s magic number mismatch detected - %p/%ly", what, magic_version, version);
   else
     api_fatal ("%s mismatch detected - %ly/%ly.\n\
-This problem is probably due to using incompatible versions of the cygwin DLL.\n\
+This problem is probably due to using incompatible versions of the msys2 or cygwin DLL.\n\
+msys2 uses pinned memory to share child information. All versions running must be the same magic_version.\n\
+\n\
+The magic versions are seen above as magic_version/version.\n\
+\n\
+Cygwin recommends the following fix:\n\
+If cygwin1.dll is installed then make sure there is only one version installed.\n\
 Search for cygwin1.dll using the Windows Start->Find/Search facility\n\
 and delete all but the most recent version.  The most recent version *should*\n\
 reside in x:\\cygwin\\bin, where 'x' is the drive on which you have\n\
 installed the cygwin distribution.  Rebooting is also suggested if you\n\
-are unable to find another cygwin DLL.",
+are unable to find another cygwin DLL.\n\
+\n\
+Since msys-2.0.dll may be installed in multiple stand-alone tools, make sure\n\
+each tool is using a compatible version of msys-2.0.dll, otherwise a mismatch\n\
+may occur. You can search for msys-2.0.dll across your system. See msys-2.0.dll\n\
+source code for history and what magic number is included.\n\
+\n\
+See https://www.msys2.org/docs/faq/ for more information.",
 	       what, magic_version, version);
 }
 


Quenty:
Sorry if this PR doesn't follow guidelines or should be submitted in another way. I couldn't find any documentation on this approach. Feel free to ignore if this isn't super useful, this is a tricky topic.

winsup/cygwin/dcrt0.cc
```diff
@@ -1360,7 +1360,10 @@ multiple_cygwin_problem (const char *what, uintptr_t magic_version, uintptr_t ve
     system_printf ("%s magic number mismatch detected - %p/%ly", what, magic_version, version);
   else
     api_fatal ("%s mismatch detected - %ly/%ly.\n\
-This problem is probably due to using incompatible versions of the cygwin DLL.\n\
+This problem is probably due to using incompatible versions of the msys2 or cygwin DLL.\n\
+msys2 is uses pinned memory to share child information. All versions running must be the same magic_version.\n\
```
jeremyd2019: this was probably supposed to be either `is using` or just `uses`.

jeremyd2019 (commented):
I'm not sure the added text really clarifies anything.  The idea of searching and destroying alternate cygwin1.dll wouldn't help as the names are different with msys2, and the idea of doing that with msys-2.0.dll is also problematic because git-for-windows' version comes from a different fork, and anyway just deleting dlls is likely to break things worse.  The best advice is just to not mix these environments, but I don't know how to word a message to that effect that would make sense.

winsup/cygwin/dcrt0.cc
```diff
@@ -1360,7 +1360,10 @@ multiple_cygwin_problem (const char *what, uintptr_t magic_version, uintptr_t ve
...
+The magic versions are seen above as magic_version/version.\n\
+If cygwin1.dll is installed then make sure there is only one version installed.\n\
```
dscho, line 1369: Since you rightfully point out that there might be an MSYS2 runtime somewhere, let's mention `msys-2.0.dll` here, too.

winsup/cygwin/dcrt0.cc
dscho: It would probably be a good idea to provide more verbose documentation about this somewhere on https://msys2.org/, and to add a link to that documentation to this here error message.

Quenty: Nice catch!

Quenty:
Ok, I've updated this with more links as well as more verbose documentation about what cygwin recommends, versus what msys2 runtime may support.

winsup/cygwin/dcrt0.cc
Quenty: Done! Other PR is pending here 

https://github.com/msys2/msys2.github.io/pull/235

winsup/cygwin/dcrt0.cc
Quenty: I didn't mention it here, because I think the advice is bad, but I did add another section underneath with better advice. Still not great, but at least it creates clarity.

dscho:
@lazka how do we want to proceed with regards to v3.4.2?

lazka:
> @lazka how do we want to proceed with regards to v3.4.2?

I've been testing it for the last few days and I'm not aware of any issues. I'll prepare a PR. edit: https://github.com/msys2/MSYS2-packages/pull/3409

lazka:
> @lazka how do we want to proceed with regards to v3.4.2?

we found another regression: https://cygwin.com/pipermail/cygwin/2022-December/252699.html

PR #127, opened 2022-12-21
[backport] Cygwin: path_conv: make sure sym.path_flags is always initialized

lazka:
Commit c1023ee353705 introduced a split between mount flags and path flags.  It didn't initialize symlink_info::path_flags in path_conv::check, because that's done in symlink_info::check.

However, there are two code paths expecting symlink_info::path_flags being already initialized and both skip symlink_info::check.

Make sure symlink_info::path_flags is initalized to 0 early in path_conv::check.

Fixes: c1023ee353705 ("Cygwin: path_conv: decouple path_types from mount types")


lazka:
maybe we should wait to 3.4.4 instead

PR #134, opened 2023-01-11
[3.3.6] Cygwin: Implicitly support the /dev/fd symlink and friends

lazka:
Bash has a very convenient feature that is called process substitution (e.g. `diff -u <(seq 0 10) <(seq 1 11)`). To make this work, Bash requires the `/dev/fd` symlink to exist, and Cygwin therefore creates this symlink (together with the `stdin`, `stdout` and `stderr` ones) upon start-up.

This strategy is incompatible with the idea of providing a subset of Cygwin in a `.zip` file (because there is no standard way to represent symlinks in `.zip` files, and besides, older Windows versions would potentially lack support for them anyway).

That type of `.zip` file is what Git for Windows wants to use, though, bundling a minimal subset for third-party applications in MinGit (see https://github.com/git-for-windows/git/wiki/MinGit for details).

Let's side-step this problem completely by creating those symlinks implicitly, similar to the way `/dev/` is populated with special devices.

Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>

### DIFF
--- a/winsup/cygwin/Makefile.am
+++ b/winsup/cygwin/Makefile.am
@@ -230,6 +230,7 @@ DLL_FILES= \
 	fhandler_console.cc \
 	fhandler_cygdrive.cc \
 	fhandler_dev.cc \
+	fhandler_dev_fd.cc \
 	fhandler_disk_file.cc \
 	fhandler_dsp.cc \
 	fhandler_fifo.cc \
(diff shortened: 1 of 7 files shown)


lazka:
I want to get rid of the fd setup code in MSYS2 (it's slow), so backport this to 3.3 too for good measure.

PR #142, opened 2023-03-29
CI: build with --disable-dependency-tracking

lazka:
We only build once in CI, so dependency tracking isn't needed. This saves 3-4 minutes in CI.

----

Not sure if that's worth it, but seemed nice.


dscho (approved):
Very nice, indeed!

PR #143, opened 2023-03-29
Stop assuming that there are no spaces in POSIX-style paths

lazka:
Git's test suite most prominently sports a POSIX path with a space in it: the tests are executed in directories whose names have the form 'trash directory.t0123-blub'. Therefore, we *must* handle those names correctly.

This fix makes Git's t1504-ceiling-dirs.sh pass.

### DIFF
--- a/winsup/cygwin/msys2_path_conv.cc
+++ b/winsup/cygwin/msys2_path_conv.cc
@@ -201,7 +201,7 @@ void ppl_convert(const char** from, const char* to, char** dst, const char* dste
 
 
 void find_end_of_posix_list(const char** to, int* in_string) {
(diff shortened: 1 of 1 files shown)


lazka:
Here is an accompanied PR adjusting the test suite: https://github.com/msys2/msys2-tests/pull/33

dscho (approved):
❤️

lazka:
There is some fallout here: https://github.com/msys2/msys2-runtime/issues/150

dscho:
> There is some fallout here: #150

😞 

I offered an analysis and a suggestion how to address the problem without breaking Git's expectations.

PR #148, opened 2023-04-03
Add a helpful debug message for posix-to-windows conversion

lazka:
https://github.com/msys2/MINGW-packages/discussions/16383

### DIFF
--- a/winsup/cygwin/msys2_path_conv.cc
+++ b/winsup/cygwin/msys2_path_conv.cc
@@ -238,6 +238,7 @@ void find_end_of_rooted_path(const char** from, const char** to, int* in_string)
 void sub_convert(const char** from, const char** to, char** dst, const char* dstend, int* in_string) {
     const char* copy_from = *from;
     path_type type = find_path_start_and_type(from, false, *to);
+    debug_printf("found type %d for path %s", type, copy_from);
 
     if (type == POSIX_PATH_LIST) {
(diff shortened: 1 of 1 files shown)


winsup/cygwin/msys2_path_conv.cc
```diff
@@ -238,6 +238,7 @@ void find_end_of_rooted_path(const char** from, const char** to, int* in_string)
 void sub_convert(const char** from, const char** to, char** dst, const char* dstend, int* in_string) {
     const char* copy_from = *from;
     path_type type = find_path_start_and_type(from, false, *to);
+    debug_printf("found type %d for path %s", type, copy_from);
```
dscho, line 241: I wonder whether it would make sense to mark this commit via `git commit --fixup 8e6ef491081215ff6f12fbe36851ed6527c384da` so that both patches are squashed together during the next rebase to a new Cygwin version?

lazka: My thoughts where that I wanted to make the range-diff easier for git4win by just keeping commits as is for a while, but fine with me.

dscho: > I wanted to make the range-diff easier for git4win

That's very kind of you!

In this instance, I don't think it is necessary, though, and I would like to focus more on keeping the MSYS2 patches in a maintainable shape.

PR #153, opened 2023-05-22
[backport] Cygwin: Fix compiling with w32api-headers v11.0.0

lazka:
This backports  https://cygwin.com/git/?p=newlib-cygwin.git;a=commit;h=1c63eaace825e32f2d03a3a228a40b1019b94c71 to fix the build with mingw-w64 v11

### DIFF
--- a/winsup/cygwin/local_includes/ntdll.h
+++ b/winsup/cygwin/local_includes/ntdll.h
@@ -518,10 +518,12 @@ enum
   FILE_RENAME_IGNORE_READONLY_ATTRIBUTE			= 0x40
 };
 
+#if (__MINGW64_VERSION_MAJOR < 11)
 enum
 {
   FILE_CS_FLAG_CASE_SENSITIVE_DIR			= 0x01
 };
+#endif
 
 enum
 {


lazka:
huh, we are still on v10 here, maybe that patch breaks the build with v10?

dscho:
Something's definitely fishy here. The [build failure](https://github.com/msys2/msys2-runtime/actions/runs/5050256343/jobs/9060738421?pr=153) indicates that indeed, `FILE_CS_FLAG_CASE_SENSITIVE_DIR` isn't declared. But the patch in this PR [makes the definition of that flag contingent on `__MINGW64_VERSION_MAJOR < 11`](https://github.com/msys2/msys2-runtime/pull/153/commits/c7cad4daa83e8be2fa1ac1c63f6f27b7e67cf0f8#diff-a1f3be1e27d6abf82903835296e6d33c563a09cc90c4bf611b3f20d5d79e1e53R521-R526).

I'm somewhat puzzled that we look at `__MINGW64_VERSION_MAJOR` when compiling with `/usr/bin/g++`, i.e. an _MSYS_ compiler? But I guess that's correct, that constant [is defined in `usr/include/w32api/_mingw_mac.h`](https://github.com/git-for-windows/git-sdk-64/blob/d034c5bae3304a13ff9f47c539993e22b05bf728/usr/include/w32api/_mingw_mac.h#L14).

Yet that set of headers does _not_ define the `FILE_CS_FLAG_CASE_SENSITIVE_DIR` flag at all!

Let's see why. That flag was added by @Biswa96 in https://sourceforge.net/p/mingw-w64/mingw-w64/ci/35fa6bdbc14383c9078520492974ff5ab31ba796. Yet in that revision, the major version was _already_ defined as 11: https://sourceforge.net/p/mingw-w64/mingw-w64/ci/35fa6bdbc14383c9078520492974ff5ab31ba796/tree/mingw-w64-headers/crt/_mingw_mac.h. But maybe that version was incremented not _just_ before releasing the new version (much like Cygwin bumps their version first thing after every release)? Indeed:

```console
$ git name-rev 35fa6bdbc14383c9078520492974ff5ab31ba796
35fa6bdbc14383c9078520492974ff5ab31ba796 tags/v11.0.0~224
```

And the [current `msys2-w32api-headers` version](https://packages.msys2.org/search?t=pkg&q=msys2-w32api-headers) is recorded as 10.0.0.r16.g49a56d453-1, and:

```console
$ git name-rev 49a56d453
49a56d453 tags/v11.0.0~257
```

Aaaaargh, so _close_! We're a scant 33 commits behind. And there is _no way_ we can tell this situation by looking at `__MINGW64_VERSION_MAJOR`.

Sadly, this patch already made it into Cygwin. I think it is faulty. A better version would be:

```diff
diff --git a/winsup/cygwin/local_includes/ntdll.h b/winsup/cygwin/local_includes/ntdll.h
index 0f2310882b..a4e8b88557 100644
--- a/winsup/cygwin/local_includes/ntdll.h
+++ b/winsup/cygwin/local_includes/ntdll.h
@@ -518,10 +518,12 @@ enum
   FILE_RENAME_IGNORE_READONLY_ATTRIBUTE			= 0x40
 };
 
+#ifndef FILE_CS_FLAG_CASE_SENSITIVE_DIR
 enum
 {
   FILE_CS_FLAG_CASE_SENSITIVE_DIR			= 0x01
 };
+#endif
 
 enum
 {
```

mati865:
> Sadly, this patch already made it into Cygwin. I think it is faulty. A better version would be:

I wouldn't say it's faulty. It could have been better, that's true but it's sensible to expect code that identifies itself as version X to have features released in version X.
The problem as you have noticed is our package that is neither v10 nor v11, it's kind of between those two.

lazka:
Thanks for the investigation.

I agree that it's not really a fault of the patch. Cygwin only packages releases and not directly from git like we do here.

I'll bump things to v11:

* https://github.com/msys2/MSYS2-packages/pull/3802
* https://github.com/msys2/MSYS2-packages/pull/3803

PR #175, opened 2023-10-29
Fix sub_convert debug output

dg0yt:
Noticed in msys2-tests repository.

### DIFF
--- a/winsup/cygwin/msys2_path_conv.cc
+++ b/winsup/cygwin/msys2_path_conv.cc
@@ -238,7 +238,7 @@ void find_end_of_rooted_path(const char** from, const char** to, int* in_string)
 void sub_convert(const char** from, const char** to, char** dst, const char* dstend, int* in_string) {
     const char* copy_from = *from;
     path_type type = find_path_start_and_type(from, false, *to);
-    debug_printf("found type %d for path %s", type, copy_from);
+    debug_printf("found type %d for path %s\n", type, copy_from);
 
     if (type == POSIX_PATH_LIST) {
         find_end_of_posix_list(to, in_string);


dscho (approved):
Obviously correct!

dscho:
@lazka do you have any idea why the CI failed?

dg0yt:
BTW I found a second statement with missing `\n`. But I may need a few days to add this.

lazka:
I think we updated gcc in the meantime.

lazka:
> I think we updated gcc in the meantime.

ah, yes , I forgot, I dealt with the same issue here: https://github.com/msys2/MSYS2-packages/commit/6cbd16b459da5e646ad2465b0d8ee64d756a9373

I will

* disable werror for now to make CI green again
* upgrade the cross compiler so that the same CFLAGS are understood by both compilers
* replace wno-error with disabling specific warnings, so we avoid hiding real errors
* hope cygwin updates their toolchain some time

lazka:
#177

dg0yt:
Is this really correct? :thinking: 
~~~
$ git grep debug_printf | grep -v '\\n' | wc -l
937
$ git grep debug_printf | grep '\\n' | wc -l
14

dscho:
> Is this really correct? 🤔
> 
> ```
> $ git grep debug_printf | grep -v '\\n' | wc -l
> 937
> $ git grep debug_printf | grep '\\n' | wc -l
> 14
> ```

Oh hahaha! I had not realized that the `debug_printf()` function (which is actually a macro that expands to another macro which eventually expands to `strace.printf()`) is backed by a function that [automatically appends a newline unless there is already one](https://github.com/msys2/msys2-runtime/blob/78f82c3df178984dc8119af3bbc82856060b0b77/winsup/cygwin/strace.cc#L192-L209)... So I guess this PR is not needed?

Note, however, that this is very different from the `small_printf()` function (with whose behavior I was confused to believe that this here PR was needed), which does distinctly [not append a newline at all](https://github.com/msys2/msys2-runtime/blob/78f82c3df178984dc8119af3bbc82856060b0b77/winsup/cygwin/smallprint.cc#L127-L354).

dg0yt:
Maybe it is easier to always add/print `\n` in msys2-tests.

PR #182, opened 2023-12-17
Backport Git for Windows' pathconv adjustments

dscho:
This PR aligns the MSYS2 runtime with the variant that Git for Windows uses. In every commit message, the test cases of Git's test suite are listed whose expectations would not be met by the MSYS2 runtime without the respective commit.

This PR is part of my effort to address https://github.com/msys2/MINGW-packages/discussions/16383.

### DIFF
--- a/newlib/libc/stdlib/mbtowc_r.c
+++ b/newlib/libc/stdlib/mbtowc_r.c
@@ -36,7 +36,7 @@ __ascii_mbtowc (struct _reent *r,
   if (n == 0)
     return -2;
 
-#ifdef __CYGWIN__
+#ifdef STRICTLY_7BIT_ASCII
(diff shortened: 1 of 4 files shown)


lazka:
lgtm

I'll try to extend our external tests with the examples provided in the commit messages once this lands.

edit: I've also confirmed that our existing tests still pass with this. (or let's say the changes behave as expected)

dscho:
I rebased this to resolve the merge conflicts.

lazka:
The test part is here: https://github.com/msys2/msys2-tests/pull/58/files

If this turns out to create any problems and there is no easy fix we could think about a `MSYS2_ARG_CONV_INCL` to opt in certain arguments to not be skipped, or something like that.

dscho:
Thank you, @lazka !

PR #184, opened 2023-12-22
fixup! CI: fix the build with gcc 13

lazka:
We've upgraded our cross compiler to gcc 13 now, so the workaround can be removed


dscho (approved):
Nice, thank you so much!

dscho:
Since this is a workflow-only change (which would be ignored by [`update-patches.sh`](https://github.com/msys2/MSYS2-packages/blob/338f3ef6f4497a056035cc20678c464e1fd14d72/msys2-runtime/update-patches.sh)), there is no need for a corresponding MSYS2-packages PR.

PR #192, opened 2024-01-30
Prevent different MSYS2 runtime versions from sharing cygheaps

dscho:
This allows users to run Cygwin programs from an MSYS2 Bash even when the MSYS2 runtime version corresponds to the Cygwin runtime version used by those Cygwin programs.

Incidentally, it also allows calling Git for Windows' Bash from the MSYS2 Bash.

### DIFF
--- a/winsup/configure.ac
+++ b/winsup/configure.ac
@@ -65,24 +65,30 @@ AC_ARG_WITH([msys2-runtime-commit],
 case "$MSYS2_RUNTIME_COMMIT" in
 no)
     MSYS2_RUNTIME_COMMIT=
+    MSYS2_RUNTIME_COMMIT_SHORT=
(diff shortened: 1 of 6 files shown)


dscho:
@lazka @Biswa96 @jeremyd2019 any thoughts whether to merge this now, or hold off until after switching to MSYS2 runtime v3.5.0?

dscho:
@Biswa96 was that a 👍 for merging now, or a 👍 for holding off until v3.5.0?

lazka:
I'd prefer to wait a bit

Biswa96:
I can not see any potential issue with this change. So, :+1: for merging it.

lazka:
I'm fine with merging this, still ok?

PR #194, opened 2024-02-01
Upgrade to Cygwin 3.5.0

dscho:
<details><summary>Range-diff relative to cygwin-3.4.10..msys2-3.4.10-autosquashed</summary>

* 1:  ea4541258a =  1:  032b7eaf84 Add MSYS2 triplet
* 2:  4fdf75c097 =  2:  b5b80682ea Fix msys library name in import libraries
* 3:  5f85b84aca !  3:  a282a19dd1 Rename dll from cygwin to msys

   ``````diff
   @@ winsup/cygwin/Makefile.am: newlib_build=$(target_builddir)/newlib
    -NEW_DLL_NAME=new-cygwin1.dll
    -DEF_FILE=cygwin.def
    -LIB_NAME=libcygwin.a
   --TEST_LIB_NAME=libcygwin0.a
    +DLL_NAME=msys-2.0.dll
    +NEW_DLL_NAME=new-msys-2.0.dll
    +DEF_FILE=msys.def
    +LIB_NAME=libmsys-2.0.a
   -+TEST_LIB_NAME=libmsys0.a
     
     #
     # sources
   @@ winsup/cygwin/Makefile.am: LIBSERVER = $(cygserver_blddir)/libcygserver.a
     $(LIBSERVER):
     	$(MAKE) -C $(cygserver_blddir) libcygserver.a
     
   --# We build as cygwin0.dll and rename at install time to overcome native
   -+# We build as msys-2.0.dll and rename at install time to overcome native
   +-# We build as new-cygwin1.dll and rename at install time to overcome native
   ++# We build as new-msys-2.0.dll and rename at install time to overcome native
     # rebuilding issues (we don't want the build tools to see a partially built
   --# cygwin.dll and attempt to use it instead of the old one).
   +-# cygwin1.dll and attempt to use it instead of the old one).
    +# msys-2.0.dll and attempt to use it instead of the old one).
     
     # linker script
   @@ winsup/cygwin/Makefile.am: LIBSERVER = $(cygserver_blddir)/libcygserver.a
     
    -# cygwin dll
    +# msys-2.0 dll
   + # Set PE and export table header timestamps to zero for reproducible builds.
     $(NEW_DLL_NAME): $(LDSCRIPT) libdll.a $(VERSION_OFILES) $(LIBSERVER)\
     		  $(newlib_build)/libm.a $(newlib_build)/libc.a
   - 	$(AM_V_CXXLD)$(CXX) $(CXXFLAGS) \
[...]

PR #196, opened 2024-02-10
pathconv: don't skip arguments with double quote

### LINKED ISSUE #190: Fallout from pathconf changes
CPython currently doesn't build because of this. Code in CPython: https://github.com/msys2-contrib/cpython-mingw/blob/ebc5f3fc250c0ae937663e43bc11ef5ab7830888/Makefile.pre.in#L1294-L1302

Before:

```
python3 -c "import sys; print(sys.argv)" -DPREFIX='"/ucrt64"'
['-c', '-DPREFIX="C:/msys64/ucrt64"']
```

After:

```
python3 -c "import sys; print(sys.argv)" -DPREFIX='"/ucrt64"'
['-c', '-DPREFIX="/ucrt64"']
```

lazka:
It is used to pass strings/paths to the preprocessor and breaks for example the CPython build. For example -DPREFIX='"/ucrt64"'.

Fixes #190

### DIFF
--- a/include/ansidecl.h
+++ b/include/ansidecl.h
@@ -283,6 +283,49 @@ So instead we use the macro below and test it against specific values.  */
 # endif /* GNUC >= 4.9 */
 #endif /* ATTRIBUTE_NO_SANITIZE_UNDEFINED */
 
+/* Attribute 'nonstring' was valid as of gcc 8.  */
+#ifndef ATTRIBUTE_NONSTRING
+# if GCC_VERSION >= 8000
+#  define ATTRIBUTE_NONSTRING __attribute__ ((__nonstring__))
+# else
+#  define ATTRIBUTE_NONSTRING
+# endif
+#endif
+
+/* Attribute `alloc_size' was valid as of gcc 4.3.  */
+#ifndef ATTRIBUTE_RESULT_SIZE_1
+# if (GCC_VERSION >= 4003)
+#  define ATTRIBUTE_RESULT_SIZE_1 __attribute__ ((alloc_size (1)))
+# else
+#  define ATTRIBUTE_RESULT_SIZE_1
+#endif
+#endif
+
+#ifndef ATTRIBUTE_RESULT_SIZE_2
+# if (GCC_VERSION >= 4003)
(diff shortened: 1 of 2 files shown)


lazka:
We are also guessing that this broke gettext relocation and the texlive build, but that isn't confirmed.

lazka:
Included the binutils fix from #194 to fix the build.

lazka:
Confirmed that this unbreaks gettext

dscho (approved):
Thank you for taking care of this. I'll have to figure out whether this causes a regression in any of the test cases listed in cfc7696e0825563fb4245377ab70887cde160cd1's commit message, but I agree that it is more important for MSYS2 to unbreak those builds.

lazka:
for future reference, this is the change for the tests: https://github.com/msys2/msys2-tests/pull/62

dscho:
> I'll have to figure out whether this causes a regression in any of the test cases listed in [cfc7696](https://github.com/msys2/msys2-runtime/commit/cfc7696e0825563fb4245377ab70887cde160cd1)'s commit message

I was able to verify that [the latest original MSYS2 runtime](https://github.com/msys2/msys2-runtime/actions/runs/7865269712#artifacts) [still allows Git's test suite to pass](https://github.com/dscho/git-for-windows-automation/actions/runs/7883924866/job/21511863639). Yay!

nyfair:
Can you simply revert 0042-fixup-Add-functionality-for-converting-UNIX-paths-in.patch?
The following characters are all valid path
```
` ' " [ ] 
```

dscho:
> Can you simply revert 0042-fixup-Add-functionality-for-converting-UNIX-paths-in.patch? The following characters are all valid path
> 
> ```
> ` ' " [ ] 
> ```

At least for double-quotes, that's not true: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions.

I do not remember why the other characters were added, but we should be able to use a variation of https://github.com/git-for-windows/git-for-windows-automation/compare/main...dscho:git-for-windows-automation:311f91570add4645b5e1babb9a013c1c4e9b2fed to see what breaks, if you happen to open a revert PR (modulo `"`).

PR #197, opened 2024-02-11
msys2-3.3.6: backports from 3.4

dscho:
As pointed out by @jeremyd2019 in https://github.com/git-for-windows/msys2-runtime/pull/61#issuecomment-1937088400, the backports I did in Git for Windows are also desirable here, even if MSYS2 is not half as reliant on v3.3.* as Git for Windows is (because it needs to [support 32-bit builds until 2029](https://gitforwindows.org/32-bit.html)).


jeremyd2019 (approved):
lgtm

PR #203, opened 2024-02-20
msys2 3.3.6 binutils fixes

### LINKED ISSUE #200: msys2-runtime-3.3 binaries no longer work
From https://github.com/msys2/MSYS2-packages/pull/4421, it was discovered that msys2-runtime-3.3-3.3.6-7 does not work, while -6 does, both on x86_64 and i686.  On i686 at least, I found that downgrading to binutils-2.40-1 resulted in a working dll while any later version I tried did not.

Again on i686, I found that it crashed early on when the dll was loaded, in a `memset` call.  It appears that the memory at `_cygheap_start` is not writable anymore.

jeremyd2019:
Fixes #200

Wound up having to backport more fixes for binutils than I expected, to get dumper.exe included again.

Corinna suggested in https://cygwin.com/pipermail/cygwin/2024-February/255472.html that there was probably no good reason for `_cygheap_start` being in `.rsrc` instead of `.cygheap` section, so moved it.  This at least works well enough to run and update other packages on both i686 and x86_64, so is definitely an improvement on the old behavior of crashing immediately.

As I mentioned in the commit message, this may result in reducing the difference between `_cygheap_start` and `_cygheap_end` by the fraction of a page not used in the `.rsrc` section.  This works out to moving `_cygheap_start` forward by 3064 bytes on i686, from 0x408 bytes into the `.rsrc` page to the start of the `.cygheap` section, which should have reduced the difference between `_cygheap_end` and `_cygheap_start` from 2137080 to 2134016 bytes.  But for some reason, in practice this difference is now 2138112 bytes.

On x86_64, the used portion of `.rsrc` is also `0x408` bytes, and the difference between end and start went from 3210232 to 3182592 bytes.

### DIFF
--- a/include/ansidecl.h
+++ b/include/ansidecl.h
@@ -283,6 +283,49 @@ So instead we use the macro below and test it against specific values.  */
 # endif /* GNUC >= 4.9 */
 #endif /* ATTRIBUTE_NO_SANITIZE_UNDEFINED */
 
+/* Attribute 'nonstring' was valid as of gcc 8.  */
+#ifndef ATTRIBUTE_NONSTRING
+# if GCC_VERSION >= 8000
+#  define ATTRIBUTE_NONSTRING __attribute__ ((__nonstring__))
+# else
+#  define ATTRIBUTE_NONSTRING
+# endif
+#endif
+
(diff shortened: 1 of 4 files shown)


lazka:
lgtm

PR #209, opened 2024-03-09
Support OneDrive better

### LINKED ISSUE #206: ls/stat on OneDrive causes download of files
If I invoke ls or anything else that does stat inside OneDrive folder it will trigger download of all files.

OneDrive uses [placeholder files](https://learn.microsoft.com/en-us/windows/win32/w8cookbook/placeholder-files) to represent remote files.

I'm guessing reading file content in stat is to support detection of actually executable files as in [here](https://github.com/msys2/msys2-runtime/blob/msys2-3.4.10/winsup/cygwin/fhandler/disk_file.cc#L548)?

I think this should be disabled on non-hydrated placeholder files.

dscho:
This backports f6b56abec1865f8da7632644210bae36b395569b and 2100c6ef0dee120c316b7bea42bf5e93c281e808, avoiding to hydrate OneDrive files (or for that matter, other files marked as "offline", i.e. not-yet-downloaded), _just_ for the sake of determining whether the files are to be considered executable by inspecting the first line in case it is `#!/bin/sh`.

This fixes #206.

### DIFF
--- a/winsup/cygwin/autoload.cc
+++ b/winsup/cygwin/autoload.cc
@@ -477,6 +477,7 @@ LoadDLLfuncEx (SetThreadDescription, 8, KernelBase, 1)
 LoadDLLfunc (VirtualAlloc2, 28, KernelBase)
 
 LoadDLLfunc (NtMapViewOfSectionEx, 36, ntdll)
+LoadDLLfuncEx (RtlSetProcessPlaceholderCompatibilityMode, 4, ntdll, 1)
 
 LoadDLLfunc (ldap_bind_s, 0, wldap32)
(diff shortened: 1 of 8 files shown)


dscho:
@mwisnicki could I ask you to test the artifact of [this PR's build](https://github.com/msys2/msys2-runtime/actions/runs/8217545455?pr=209#artifacts) (i.e. unpack the `usr\bin\msys-2.0.dll` from that `install.zip` over your MSYS2 installation's, then try `ls` with a OneDrive folder)?

mwisnicki:
Works correctly. Thank you.

Issue #208, opened 2024-03-08
Regression from pathconv changes


naveen521kk:
Previously msys2-runtime converted the path separator from `:` to `;` when passing to a Windows program. Now, for some reason, this doesn't work:
```
$ A=.:../test python -c 'import os; print(os.environ["A"])'
.:../test
```
Though if it's the `PATH` variable it seems to work.
```
$ PATH=".:../test" /mingw64/bin/python -c 'import os; print(os.environ["PATH"])'
.;C:\Users\test
```

Output from an old version of runtime for reference:
```
$ A=.:../test python -c 'import os; print(os.environ["A"])'
.;..\test
```

This is probably why rebuilding texlive errors out, see https://github.com/msys2/MINGW-packages/pull/19921#issuecomment-1923454422.

Voxeles:
Seems like this is caused by https://github.com/msys2/msys2-runtime/commit/9944652c41fb96f555d221865f4ed5d0c8b33514
Any path with `:.` or `:..` in it is ignored.

For example, this works:
```
$ A="/this:/there" python -c 'import os; print(os.environ["A"])'
C:\msys64\this;C:\msys64\there
$ A="./this:/there" python -c 'import os; print(os.environ["A"])'
.\this;C:\msys64\there
``` 

But this doesn't:
```
A="/this:./there" python -c 'import os; print(os.environ["A"])'
/this:./there
A="./this:./there" python -c 'import os; print(os.environ["A"])'
./this:./there
```

reneparis:
@dscho It seems as your commit https://github.com/msys2/msys2-runtime/commit/9944652c41fb96f555d221865f4ed5d0c8b33514 broke the pathconv rules as described [here](https://www.msys2.org/docs/filesystem-paths/). Unfortunately, the rules are not complete w.r.t. relative paths. The official documentation says:

```bash
$ python3 -c "import sys; print(sys.argv)" --dir=/foo
['-c', '--dir=C:/msys64/foo']
$ python3 -c "import sys; print(sys.argv)" --dir=/foo:/bla
['-c', '--dir=C:\\msys64\\foo;C:\\msys64\\bla']
```
Relative paths have not been mentioned so far, but the expected behavior should be something like this:

```bash
$ python3 -c "import sys; print(sys.argv)" --dir=/foo:./bla
['-c', '--dir=C:\\msys64\\foo;.\\bla']
```

With the current pathconv rules, we get a different - flawed - picture

```bash
[...]

PR #212, opened 2024-04-17
Recognize path lists starting with absolute or relative paths again

dscho:
When we fixed MSYS2's automatic Unix <-> Windows path conversion to identify and skip Git-style `<rev>:<path>` arguments (and incidentally also scp-style `<host>:<path>` ones), we assumed that path lists containing relative paths would be a rare scenario.

My, was this assumption wrong!

Let's add another heuristic that detects absolute paths at the beginning of path lists, and relative ones starting with either `./` or `../`, neither of which match those Git-style nor scp-style arguments, and then prevent the detection of the latter style from kicking in.

This addresses https://github.com/msys2/msys2-runtime/issues/208

### DIFF
--- a/winsup/cygwin/msys2_path_conv.cc
+++ b/winsup/cygwin/msys2_path_conv.cc
@@ -359,6 +359,18 @@ skip_p2w:
     }
 
     /*
+     * Discern between Git's `<rev>:<path>`, SCP's `<host>:<path>` pattern
+     * (which is not a path list but may naïvely look like one) on the one
+     * hand, and path lists starting with `/<path>`, `./<path>` or `../<path>`
+     * on the other hand.
+     */
+    bool potential_path_list = *it == '/' ||
+	    (*it == '.' &&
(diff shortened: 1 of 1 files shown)


reneparis:
Just integrated the build artifacts 😞 

:x:`python3 -c "import sys; print(sys.argv)" --dir=/this:./there`
- Actual:  `['-c', '--dir=/this:./there']`
- Expected: `['-c', '--dir=C:\\msys64\\this;.\\there']`  

:x:`python3 -c "import sys; print(sys.argv)" --dir=./this:./there`
- Actual:  `['-c', '--dir=./this:./there']`
- Expected: `['-c', '--dir=.\\this;.\\there']`  

:x:`python3 -c "import sys; print(sys.argv)" --WEBINPUTS=.:../../../texk/web2c`
- Actual: `['-c', '--WEBINPUTS=.:../../../texk/web2c']`
- Expected: `['-c', '--WEBINPUTS=.;..\\..\\..\\texk\\web2c']`

PR #213, opened 2024-04-29
Fix pathconv issues related to relative paths in path lists

ReinhardBiegelIntech:
Implementation associated with testsuite changes here: https://github.com/msys2/msys2-tests/pull/67

Only tested in testsuite yet.

@dscho

### DIFF
--- a/winsup/cygwin/msys2_path_conv.cc
+++ b/winsup/cygwin/msys2_path_conv.cc
@@ -359,6 +359,18 @@ skip_p2w:
     }
 
     /*
+     * Discern between Git's `<rev>:<path>`, SCP's `<host>:<path>` pattern
+     * (which is not a path list but may naïvely look like one) on the one
+     * hand, and path lists starting with `/<path>`, `./<path>` or `../<path>`
+     * on the other hand.
+     */
+    bool potential_path_list = *it == '/' ||
+	    (*it == '.' &&
+		(it[1] == ':' || it[1] == '/' ||
+		 (it[1] == '.' &&
+			(it[2] == ':' || it[2] == '/'))));
+
+    /*
      * Prevent Git's :file.txt and :/message syntax from beeing modified.
      */
     if (*it == ':')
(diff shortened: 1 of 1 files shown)


reneparis:
@lazka

winsup/cygwin/msys2_path_conv.cc
```diff
@@ -358,6 +358,40 @@ path_type find_path_start_and_type(const char** src, int recurse, const char* en
...
+     * (which is not a path list but may naïvely look like one) on the one
+     * hand, and path lists starting with `/<path>`, `./<path>` or `../<path>`
+     * on the other hand.
+     */
+    bool potential_path_list = *it == '/' ||
+	    (*it == '.' &&
+		(it[1] == ':' || it[1] == '/' ||
+		 (it[1] == '.' &&
+			(it[2] == ':' || it[2] == '/'))));
+
```
bustercopley:
```suggestion
    /*
     * Discern between Git's `<rev>:<path>`, SCP's `<host>:<path>` pattern
     * (which is not a path list but may naively look like one) on the one
     * hand, and path lists starting with `/<path>`, `./<path>` or `../<path>`
     * on the other hand.
     */
    bool potential_path_list = *it == '/' ||
        (*it == '.' &&
         (it + 1 == end || it[1] == '\0' || it[1] == ':' || it[1] == '/' ||
          (it[1] == '.' &&
           (it + 2 == end || it[2] == '\0' || it[2] == ':' || it[2] == '/'))));

```
* Watch for end-of-string / end of input.
* This file mostly doesn't use tabs.
* Let's not introduce a character `ï` that isn't encoded the same in iso-latin1 and utf-8 (there are no others in this file yet).

bustercopley (changes requested):

winsup/cygwin/msys2_path_conv.cc
bustercopley: @reneparis, what do you think?

bustercopley: @ReinhardBiegelIntech, what do you think?

(This replaces a deleted comment which tagged @ reneparis. Apologies, René.)

reneparis: I checked your code and agree on your suggestions - so let's wait for reinhard (currently on vacation) to implement it as soon as he's back.

ReinhardBiegelIntech: Thanks for your feedback, @bustercopley. Suggestions look good to me, so I updated the PR accordingly. Also committed to [test suite](https://github.com/msys2/msys2-tests/pull/67).

As the changes affect code originally written bei @dscho, I'd also like to get his feedback here.

ReinhardBiegelIntech:
Seems like this doesn't match any maintainer's idea of how to address #208, so im closing here.

PR #227, opened 2024-09-03
Cygwin: pipe: Fix a regression that raw_write() slows down

dscho:
It was [reported to the Cygwin mailing list](https://cygwin.com/pipermail/cygwin/2024-August/256398.html) that a `grep` on a large file was much slower than it used to be. This is a backport of the fix.


lazka:
lgtm

PR #234, opened 2024-11-13
cygthread: suspend thread before terminating.

jeremyd2019:
This addresses an extremely difficult to debug deadlock when running under emulation on ARM64.

A relatively easy way to trigger this bug is to call `fork()`, then within the child process immediately call another `fork()` and then `exit()` the intermediate process.

It would seem that there is a "code emulation" lock on the wait thread at this stage, and if the thread is terminated too early, that lock still exists albeit without a thread, and nothing moves forward.

It seems that a `SuspendThread()` combined with a `GetThreadContext()` (to force the thread to _actually_ be suspended, for more details see https://devblogs.microsoft.com/oldnewthing/20150205-00/?p=44743) makes sure the thread is "booted" from emulation before it is suspended.

Hopefully this means it won't be holding any locks or otherwise leave emulation in a bad state when the thread is terminated.

Also, attempt to use `CancelSynchonousIo()` (as seen in `flock.cc`) to avoid the need for `TerminateThread()` altogether.  This doesn't always work, however, so was not a complete fix for the deadlock issue.

Addresses: https://cygwin.com/pipermail/cygwin-developers/2024-May/012694.html

Fixes msys2/msys2-autobuild#62, fixes #228 (and some other issues scattered about I don't remember off-hand)

Awaiting review on cygwin-patches mailing list:  https://inbox.sourceware.org/cygwin-patches/2c68d6fe-5493-b7e0-6335-de5a68d3cd3f@jdrake.com/T/#u

PR #233, opened 2024-11-11
Cygwin: find_fast_cwd: don't run assembler checking code on ARM64

dscho:
The code that looks for the `FAST_CWD` pointer seems to crash a lot in recent Canary builds of Windows. This has been fixed in a commit on Cygwin's (yet unreleased) 3.6.0 branch, in 4e77fa9b8b (Cygwin: find_fast_cwd: don't run assembler checking code on ARM64, 2024-02-13), and I already backported this into Git for Windows' fork of the MSYS2 runtime.

Let's backport it to MSYS2 proper, too.

This fixes https://github.com/msys2/msys2-runtime/issues/232

PR #235, opened 2024-11-13
backport arm64 fixes for msys2-3_3_6-release

jeremyd2019:
Since I already did this for testing on Windows 10 ARM64, I thought I'd open a PR for these too.  This is effectively #233 and #234

### DIFF
--- a/winsup/cygwin/cygthread.cc
+++ b/winsup/cygwin/cygthread.cc
@@ -302,6 +302,20 @@ cygthread::terminate_thread ()
   if (!inuse)
     goto force_notterminated;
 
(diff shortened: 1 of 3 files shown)


jeremyd2019:
I don't use msys2-runtime-3.4, and don't see any use case for it on arm64, so I haven't tried to backport anything to it.

lazka:
-> https://github.com/msys2/MSYS2-packages/pull/5006

dscho:
Does this also need #236?

jeremyd2019:
yes.  I was waiting for confirmation that it fixed the issue.

PR #236, opened 2024-11-13
fixup! cygthread: suspend thread before terminating.

jeremyd2019:
Suppress error output if ReadFile on child wait pipe returns ERROR_OPERATION_ABORTED due to addition of CancelSynchronousIo call.

@lazka can you test this, since you saw the error?

### DIFF
--- a/winsup/cygwin/pinfo.cc
+++ b/winsup/cygwin/pinfo.cc
@@ -1262,13 +1262,14 @@ proc_waiter (void *arg)
 
   for (;;)
     {
-      DWORD nb;
+      DWORD nb, err;
       char buf = '\0';
 
       if (!ReadFile (vchild.rd_proc_pipe, &buf, 1, &nb, NULL)
-	  && GetLastError () != ERROR_BROKEN_PIPE)
+	  && (err = GetLastError ()) != ERROR_BROKEN_PIPE)
 	{
-	  system_printf ("error on read of child wait pipe %p, %E", vchild.rd_proc_pipe);
+	  if (err != ERROR_OPERATION_ABORTED)
+	    system_printf ("error on read of child wait pipe %p, %E", vchild.rd_proc_pipe);
 	  break;
 	}
 


jeremyd2019:
oops, should have read closer...

winsup/cygwin/pinfo.cc
```diff
@@ -1268,7 +1268,8 @@ proc_waiter (void *arg)
       if (!ReadFile (vchild.rd_proc_pipe, &buf, 1, &nb, NULL)
 	  && GetLastError () != ERROR_BROKEN_PIPE)
 	{
-	  system_printf ("error on read of child wait pipe %p, %E", vchild.rd_proc_pipe);
+	  if (GetLastError() != ERROR_OPERATION_ABORTED)
```
jeremyd2019: I don't particularly like calling `GetLastError` twice in a row like that (really, 3 times since apparently `%E` in system_printf does so also), but any other approach I could think of was less clean.

lazka: just wondering, so we want all the process exit logic below to be skipped in case of ERROR_OPERATION_ABORTED?

jeremyd2019: yes, if we called `CancelSynchronousIo` it's because we were going to terminate the thread before, so we want it to bail out pretty much as quickly as possible because the process is being torn down.

It took me a bit to wrap my head around why we *didn't* want to skip that logic in case of `ERROR_BROKEN_PIPE`: it's because the other end of the pipe is held by the child process, and buf is initialized to `'\0'`, so it will treat that as child exited in the switch below.

jeremyd2019: ok, couldn't help myself 😉

lazka: ok, thanks

lazka:
For future reference, this was the issue this is fixing: https://github.com/msys2/msys2-runtime/pull/234#issuecomment-2474986546

jeremyd2019:
https://inbox.sourceware.org/cygwin-patches/ac88704b-3f63-1f14-3412-4acce012f729@jdrake.com/T/#u

PR #239, opened 2024-11-20
fixup cygthread: suspend thread before terminating to version merged upstream.

jeremyd2019:
Address review comments:
* add comments
* change nested ifs to && conditions

Added CancelSynchronousIo before another terminate_thread call, which was not the one that was causing issues, but still makes sense to avoid termination if possible.

This brings the whole thing (after `autosquash`) to match what was merged upstream (b091b47b9e569389ad68bf28274c88d22be38ca5).

Should I add some sort of "backported from" trailer to the commit message too (using `--fixup=reword:`?)  Or is there some other git trick I should use to make sure this works out clearly when rebased in future versions?

### DIFF
--- a/winsup/cygwin/pinfo.cc
+++ b/winsup/cygwin/pinfo.cc
@@ -1268,6 +1268,9 @@ proc_waiter (void *arg)
       if (!ReadFile (vchild.rd_proc_pipe, &buf, 1, &nb, NULL)
 	  && (err = GetLastError ()) != ERROR_BROKEN_PIPE)
 	{
+	  /* ERROR_OPERATION_ABORTED is expected due to the possibility that
+	     CancelSynchronousIo interruped the ReadFile call, so don't output
+	     that error */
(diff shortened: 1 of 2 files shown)


dscho (approved):
Excellent!

dscho:
> Should I add some sort of "backported from" trailer to the commit message too (using `--fixup=reword:`?) Or is there some other git trick I should use to make sure this works out clearly when rebased in future versions?

I think you already did the best: mark it as a `fixup!`. That way, we just need to remember to rebase to the next Cygwin version using _two_ rebases: a first one with base `cygwin-3.5.4` to auto-squash the `fixup!`, and then a second one onto the new Cygwin version (which will either skip the now-auto-squashed because it looks identical to the upstream version, or it will stop with a "No changes" warning if the diff context is too different for `git rev-list --cherry-pick` to identify it as identical).

dscho:
⇉ https://github.com/msys2/MSYS2-packages/pull/5015

PR #241, opened 2024-11-21
backport fixups to 3.3.6

jeremyd2019:


jeremyd2019:
maybe hold off until we know for sure what's going on with the hang report in msys2/MSYS2-packages#4340

PR #243, opened 2024-11-22
Cygwin: revert use of CancelSyncronousIo on wait_thread.

jeremyd2019:
It appears this is causing hangs on native x86_64 in similar scenarios
as the hangs on ARM64, because `CancelSynchronousIo` is returning `TRUE`
but not canceling the `ReadFile` call as expected.

Addresses: https://github.com/msys2/MSYS2-packages/issues/4340#issuecomment-2491401847
Fixes: b091b47b9e56 ("cygthread: suspend thread before terminating.")

created via `git checkout 16dfe943bcba396ac0b11c5b52cdbaeed4e03a8d -- sigproc.cc pinfo.cc`

### DIFF
--- a/winsup/cygwin/pinfo.cc
+++ b/winsup/cygwin/pinfo.cc
@@ -1262,17 +1262,13 @@ proc_waiter (void *arg)
 
   for (;;)
     {
-      DWORD nb, err;
+      DWORD nb;
       char buf = '\0';
 
       if (!ReadFile (vchild.rd_proc_pipe, &buf, 1, &nb, NULL)
-	  && (err = GetLastError ()) != ERROR_BROKEN_PIPE)
+	  && GetLastError () != ERROR_BROKEN_PIPE)
 	{
-	  /* ERROR_OPERATION_ABORTED is expected due to the possibility that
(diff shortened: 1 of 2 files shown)


dscho (approved):
Changes look good to me; I'd love some more details in the commit message, but won't insist on them (following the link provides some context, it's just that it relies on a working internet/GitHub).

jeremyd2019:
This was a quick and dirty commit for testing, in response to https://github.com/msys2/MSYS2-packages/issues/4340#issuecomment-2492733866 - I fired up some tests to run while I slept, and they seem to not have hung on either x86_64 or ARM64, so I can come back to this and clean up the commit message and submit to cygwin-patches once approved here

dscho (approved):
Makes sense to me!

dscho:
↝https://github.com/msys2/MSYS2-packages/pull/5019

jeremyd2019:
https://inbox.sourceware.org/cygwin-patches/65158100-4a68-30c8-8060-e8fef1861c5d@jdrake.com/T/#u

jeremyd2019:
6614da9a71fc5cf24991b47fd62a56749c7f4ccc

PR #244, opened 2024-11-29
Cygwin: uname: add host machine tag to sysname

### LINKED ISSUE #238: identify arm64 host in uname output?
I was recently thinking about how `uname` used to include `-WOW64` for i686-on-x86_64 situations, and that it would be possible to use `IsWow64Process2` to identify and report x86_64-on-ARM64 (and indeed i686-on-ARM64, if backported to 3.3) in a similar way.  That might be useful for stats collection purposes (such as https://github.com/msys2/MSYS2-packages/discussions/2832).

I thought to bring this up on cygwin list, but I already have several outstanding messages there that have never been replied to, and I'd much rather my recent patch for hangs on ARM64 get attention, so I figured I'd memorialize the idea here.

jeremyd2019:
This backports two commits from cygwin master, caching IsWow64Process2 and the cygwin module's architecture in wincap, and using this to add a host machine tag to uname's sysname field, like used to be done with `-WOW64` when that was supported (see forthcoming PR to 3.3.6 for more complete version that includes handling of i686 cases, which were removed during review before applying to cygwin, since they no longer support i686)

Closes #238

### DIFF
--- a/winsup/cygwin/local_includes/wincap.h
+++ b/winsup/cygwin/local_includes/wincap.h
@@ -42,6 +42,8 @@ class wincapc
   RTL_OSVERSIONINFOEXW	version;
   char			osnam[40];
   const void		*caps;
+  USHORT		host_mach;
+  USHORT		cygwin_mach;
   bool			_is_server;
 
 public:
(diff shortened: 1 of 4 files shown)


dscho (approved):

dscho:
➡️https://github.com/msys2/MSYS2-packages/pull/5035

PR #252, opened 2025-01-16
Some `msys2-runtime-commit` fixes

dscho:
I noticed recently that the `msys-2.0.dll` files built in PR builds lacked the commit information, which makes it easier to get confused when bisecting issues such as the v3.5.5 hangs.

This PR does not fix the hangs, but the recorded commit information (that is part of the output of `uname -a`).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,7 +24,7 @@ jobs:
           # resulting in errors due to -Werror. Disable them for now.
           export CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=array-bounds -Wno-error=overloaded-virtual -Wno-narrowing -Wno-use-after-free -Wno-error=maybe-uninitialized"
           (cd winsup && ./autogen.sh)
-          ./configure --disable-dependency-tracking
+          ./configure --disable-dependency-tracking --with-msys2-runtime-commit=$GITHUB_SHA
           make -j8
 
       - name: Install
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
           # resulting in errors due to -Werror. Disable them for now.
           export CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=array-bounds -Wno-error=overloaded-virtual -Wno-narrowing -Wno-use-after-free -Wno-error=maybe-uninitialized"
           (cd winsup && ./autogen.sh)
-          ./configure --disable-dependency-tracking
+          ./configure --disable-dependency-tracking --with-msys2-runtime-commit=$GITHUB_SHA
```
jeremyd2019:
```suggestion
          ./configure --disable-dependency-tracking --with-msys2-runtime-commit="$GITHUB_SHA"
```

just in case?

jeremyd2019 (approved):

.github/workflows/build.yaml
dscho: Thank you for the suggestion, and yes: I agree that this is a good idea, in particular in the light of [the recent Python incident](https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection).

lazka:
lgtm

.github/workflows/build.yaml
mati865: Time to add `zizmor` so CI can lint our CI? 😉

dscho: > Time to add `zizmor` so CI can lint our CI? 😉

@mati865 go for it!

dscho:
I won't open a corresponding MSYS2-packages PR because this PR does not result in any functional changes of the `msys2-runtime` package, it is relevant only for future PRs.

PR #254, opened 2025-01-26
Rebase to Cygwin v3.5.6

dscho:
<details><summary>Range-diff to v3.5.5</summary>

* 1:  bdad035f0b =  1:  61c7294e59 Add MSYS2 triplet
* 2:  900232bfff =  2:  e6895db114 Fix msys library name in import libraries
* 3:  585fadf7ea =  3:  e501600b6d Rename dll from cygwin to msys
* 4:  d076660c2d =  4:  70cdcbadb9 Add functionality for converting UNIX paths in arguments and environment variables to Windows form for native Win32 applications.
* 5:  35a477b5cd =  5:  d6cd58e21e Add functionality for changing OS name via MSYSTEM environment variables.
* 6:  41ed62e6f5 =  6:  e7d3c8f96e - Move root to /usr. - Change sorting mount points. - By default mount without ACLs. - Can read /etc/fstab with short mount point format.
* 7:  8096583fac =  7:  8c5cc2861a Instead of creating Cygwin symlinks, use deep copy by default
* 8:  0ba54e8cc1 =  8:  e9074291e7 Automatically rewrite TERM=msys to TERM=cygwin
* 9:  e28b7422e2 =  9:  8102cded4c Do not convert environment for strace
* 10:  4a24304b8b = 10:  c68f8fd84b strace.cc: Don't set MSYS=noglob
* 11:  b3a4a5bd44 = 11:  e6cc70d89e Add debugging for strace make_command_line
* 12:  f8c098028c = 12:  e25ab8660e strace --quiet: be *really* quiet
* 13:  c4d9d83ede = 13:  8e182331c4 path_conv: special-case root directory to have trailing slash
* 14:  1c5a6e869c = 14:  0c102ea21b When converting to a Unix path, avoid double trailing slashes
* 15:  8b5e9b55a1 = 15:  8ed6549f80 msys2_path_conv: pass PC_NOFULL to path_conv
* 16:  b09fb7fd5a = 16:  853cc5411c path-conversion: Introduce ability to switch off conversion.
* 17:  b2391806db = 17:  5d54a247cd dcrt0.cc: Untangle allow_glob from winshell
* 18:  1e94f3ee1d = 18:  d85da3a918 dcrt0.cc (globify): Don't quote literal strings differently when dos_spec
* 19:  e6c5e3ec30 = 19:  a3e7629f84 Add debugging for build_argv
* 20:  d44fc10030 = 20:  ca226ec37b environ.cc: New facility/environment variable MSYS2_ENV_CONV_EXCL
* 21:  98c164175f = 21:  89f8cacab1 Fix native symbolic link spawn passing wrong arg0
* 22:  d12ab6d59c = 22:  a46c7539a1 Introduce the `enable_pcon` value for `MSYS`
* 23:  865ad5d64e = 23:  5f0df244ed popen: call /usr/bin/sh instead of /bin/sh
* 24:  5f36fdbfcd = 24:  c1a941081d Disable the 'cygwin' GitHub workflow
* 25:  862887f29f ! 25:  93544235a4 CI: add a GHA for doing a basic build test

   ``````diff
   @@ .github/workflows/build.yaml (new)
    +        shell: msys2 {0}
    +        run: |
    +          (cd winsup && ./autogen.sh)
   -+          ./configure --disable-dependency-tracking
   ++          ./configure --disable-dependency-tracking --with-msys2-runtime-commit="$GITHUB_SHA"
    +          make -j8
    +
    +      - name: Install
   ``````

* 26:  764bd7fb8f ! 26:  3edbe01946 CI: fix the build with gcc 13

   ``````diff
   @@ .github/workflows/build.yaml: jobs:
    +          # resulting in errors due to -Werror. Disable them for now.
    +          export CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=array-bounds -Wno-error=overloaded-virtual -Wno-narrowing -Wno-use-after-free"
               (cd winsup && ./autogen.sh)
   -           ./configure --disable-dependency-tracking
   +           ./configure --disable-dependency-tracking --with-msys2-runtime-commit="$GITHUB_SHA"
               make -j8
   ``````

* 27:  bea6a078ca = 27:  d49861d9ab Set up a GitHub Action to keep in sync with Cygwin
* 28:  080e2b0f6a = 28:  6d0c6370ed Expose full command-lines to other Win32 processes by default
* 29:  9f45f58388 = 29:  0e5a7de4fc Add a helper to obtain a function's address in kernel32.dll
* 30:  f6f7e2c928 = 30:  98632e154d Emulate GenerateConsoleCtrlEvent() upon Ctrl+C
* 31:  86cd50d151 = 31:  c890b0cc16 kill: kill Win32 processes more gently
* 32:  4b228ef2e4 = 32:  5ab3720b4f Cygwin: make option for native inner link handling.
* 33:  6e0c3b56c7 = 33:  1fcaaa3ff2 docs: skip building texinfo and PDF files
* 34:  ecf50e64ce = 34:  bbd5897e69 install-libs: depend on the "toollibs"
* 35:  8cb4ff2c0e = 35:  7e40bdfb02 POSIX-ify the SHELL variable
* 36:  1f6020c6bf = 36:  4c80b59ea3 Handle ORIGINAL_PATH just like PATH
* 37:  79e3880244 = 37:  0c124b9088 uname: allow setting the system name to CYGWIN
* 38:  ef59c45434 = 38:  54d1ac996b Pass environment variables with empty values
* 39:  c347d6c26b = 39:  6fe00e22de Optionally disallow empty environment values again
* 40:  210e5f3c4d = 40:  82d1e3b556 build_env(): respect the `MSYS` environment variable
* 41:  bef1e6ba82 = 41:  855f33f4b9 Revert "Cygwin: Enable dynamicbase on the Cygwin DLL by default"
* 42:  8103a099cf ! 42:  c8c911a222 CI: set -Wno-error=maybe-uninitialized

   ``````diff
   @@ .github/workflows/build.yaml: jobs:
    -          export CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=array-bounds -Wno-error=overloaded-virtual -Wno-narrowing -Wno-use-after-free"
    +          export CXXFLAGS="-Wno-error=stringop-truncation -Wno-error=array-bounds -Wno-error=overloaded-virtual -Wno-narrowing -Wno-use-after-free -Wno-error=maybe-uninitialized"
               (cd winsup && ./autogen.sh)
   -           ./configure --disable-dependency-tracking
   +           ./configure --disable-dependency-tracking --with-msys2-runtime-commit="$GITHUB_SHA"
               make -j8
   ``````

* 43:  6e10b1bdd9 ! 43:  f840e12ed7 Avoid sharing cygheaps across Cygwin versions

   ``````diff
   @@ winsup/configure.ac: AC_CHECK_TOOL(RANLIB, ranlib, ranlib)
    +    then
    +        MSYS2_RUNTIME_COMMIT_HEX="0x$(expr "$MSYS2_RUNTIME_COMMIT" : '\(.\{,8\}\)')ull"
    +    else
   -+        AC_MSG_WARN([Could not determine msys2-runtime commit"])
   ++        AC_MSG_WARN([Could not determine msys2-runtime commit])
    +        MSYS2_RUNTIME_COMMIT=
    +        MSYS2_RUNTIME_COMMIT_HEX=0
    +    fi
   ``````

* 44:  f61f7fb71a ! 44:  905cab18b5 uname: report msys2-runtime commit hash, too

   ``````diff
   @@ winsup/configure.ac: AC_ARG_WITH([msys2-runtime-commit],
    +        MSYS2_RUNTIME_COMMIT_SHORT="$(expr "$MSYS2_RUNTIME_COMMIT" : '\(.\{,8\}\)')"
    +        MSYS2_RUNTIME_COMMIT_HEX="0x${MSYS2_RUNTIME_COMMIT_SHORT}ul"
         else
   -         AC_MSG_WARN([Could not determine msys2-runtime commit"])
   +         AC_MSG_WARN([Could not determine msys2-runtime commit])
             MSYS2_RUNTIME_COMMIT=
    +        MSYS2_RUNTIME_COMMIT_SHORT=
             MSYS2_RUNTIME_COMMIT_HEX=0
   ``````

* 45:  debafbf93f = 45:  856e949833 Cygwin: find_fast_cwd: don't run assembler checking code on ARM64
* 46:  2c55ca50ee = 46:  c94b46ba03 cygthread: suspend thread before terminating.
* 47:  e5dc132b3f = 47:  62f6088206 Cygwin: revert use of CancelSyncronousIo on wait_thread.
* 48:  5453f9fab6 = 48:  7f3ae80791 Cygwin: cache IsWow64Process2 host arch in wincap.
* 49:  985e2654b8 = 49:  7da06fd0ea Cygwin: uname: add host machine tag to sysname.
* 50:  33799c250f <  -:  ---------- Cygwin: signal: Do not handle signal when __SIGFLUSHFAST is sent
* 51:  964efe3b52 <  -:  ---------- fixup! CI: add a GHA for doing a basic build test
* 52:  25700873e4 <  -:  ---------- fixup! Avoid sharing cygheaps across Cygwin versions
* 53:  bfb6ba0ad0 <  -:  ---------- fixup! CI: add a GHA for doing a basic build test
* 54:  674137e51c <  -:  ---------- Revert "Cygwin: signal: Do not handle signal when __SIGFLUSHFAST is sent"
* 55:  16b83734da <  -:  ---------- Cygwin: cygwait: Make cygwait() reentrant
* 56:  ae26e1742a <  -:  ---------- Cygwin: signal: Do not handle signal when __SIGFLUSHFAST is sent
* 57:  25a3862d02 <  -:  ---------- Cygwin: signal: Avoid frequent TLS lock/unlock for SIGCONT processing
*  -:  ---------- > 50:  2c0e426da7 fixup! Instead of creating Cygwin symlinks, use deep copy by default

</details>


winsup/cygwin/path.cc
```diff
@@ -2220,6 +2311,77 @@ symlink_worker (const char *oldpath, path_conv &win32_newpath, bool isdevice)
...
+          if (wsym_type == WSYM_deepcopy)
+	    {
+	      path_conv src_path;
+	      src_path.check (oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
+	      if (src_path.error)
+		{
+		  set_errno (src_path.error);
+		  __leave;
+		}
+	      if (!src_path.isondisk () && !src_path.is_fs_special ())
```
jeremyd2019: I don't think that's right... perhaps
```suggestion
	      if (src_path.isondisk ())
```

dscho:
> * 46:  [2c55ca5](https://github.com/msys2/msys2-runtime/commit/2c55ca50eeb8532432f51391c2ec28a14fc4e510) = 46:  [c94b46b](https://github.com/msys2/msys2-runtime/commit/c94b46ba0328dcba25f6a541f1c60ab53c06754e) cygthread: suspend thread before terminating.
> 
> * 47:  [e5dc132](https://github.com/msys2/msys2-runtime/commit/e5dc132b3fa1c6a06867e73c7f67c81a502e3993) = 47:  [62f6088](https://github.com/msys2/msys2-runtime/commit/62f6088206fcf4c9cbd74b23a869b7547f640c65) Cygwin: revert use of CancelSyncronousIo on wait_thread.

@jeremyd2019 do we want to drop these?

winsup/cygwin/path.cc
jeremyd2019: I don't know for sure, I was trying to reason about the original, and then started trying to reason about what you changed it to... so I'm all sorts of mixed up on what it's trying to do atm :grin:

jeremyd2019:
> > * 46:  [2c55ca5](https://github.com/msys2/msys2-runtime/commit/2c55ca50eeb8532432f51391c2ec28a14fc4e510) = 46:  [c94b46b](https://github.com/msys2/msys2-runtime/commit/c94b46ba0328dcba25f6a541f1c60ab53c06754e) cygthread: suspend thread before terminating.
> > * 47:  [e5dc132](https://github.com/msys2/msys2-runtime/commit/e5dc132b3fa1c6a06867e73c7f67c81a502e3993) = 47:  [62f6088](https://github.com/msys2/msys2-runtime/commit/62f6088206fcf4c9cbd74b23a869b7547f640c65) Cygwin: revert use of CancelSyncronousIo on wait_thread.
> 
> @jeremyd2019 do we want to drop these?

No, these are the fixes for the arm64 hang, they did not get backported to cygwin-3_5-branch.  Maybe squash them into one if you really want to, but they went into cygwin/master like that...

dscho:
> > > * 46:  [2c55ca5](https://github.com/msys2/msys2-runtime/commit/2c55ca50eeb8532432f51391c2ec28a14fc4e510) = 46:  [c94b46b](https://github.com/msys2/msys2-runtime/commit/c94b46ba0328dcba25f6a541f1c60ab53c06754e) cygthread: suspend thread before terminating.
> > > * 47:  [e5dc132](https://github.com/msys2/msys2-runtime/commit/e5dc132b3fa1c6a06867e73c7f67c81a502e3993) = 47:  [62f6088](https://github.com/msys2/msys2-runtime/commit/62f6088206fcf4c9cbd74b23a869b7547f640c65) Cygwin: revert use of CancelSyncronousIo on wait_thread.
> > 
> > 
> > @jeremyd2019 do we want to drop these?
> 
> No, these are the fixes for the arm64 hang, they did not get backported to cygwin-3_5-branch.

Ah, sorry, I thought that the actual fixes went into `cygwin-3_5-branch` and that these were some extra left-overs from earlier attempts. Thank you for clarifying!

> Maybe squash them into one if you really want to, but they went into cygwin/master like that...

Nah, better keep them as-are, so that the rebase to v3.6 will be able to figure out automatically that they are upstream already.

winsup/cygwin/path.cc
jeremyd2019: the original was
```c++
if (!src_path.isdevice () && !src_path.is_fs_special ())
```

isdevice was
```c++
return dev.not_device (FH_FS) && dev.not_device (FH_FIFO);
```
where not_device is
```c++
return d.devn && n != d.devn;
```

is_fs_special is
```c++
return dev_on_fs && d.devn != FH_FS;
```


so expanded that would be
```c++
if ((!d.devn || d.devn == FH_FS || !d.devn || d.devn == FH_FIFO) && (!dev_on_fs || d.devn == FH_FS))
```

the new isondisk is
```c++
dev_on_fs || d.devn == FH_FS
```

so I'm pretty sure `isondisk ()` is roughly equivalent, with the possible exception of devn == 0 or devn == FH_FIFO.

dscho:
Uh oh. Git's test suite [is not happy](https://github.com/git-for-windows/msys2-runtime/actions/runs/12977826426/job/36191793757?pr=84):

image

I suspect that my `fixup!`s are incorrect because those failures seem to occur around symlinks, which the `fixup!` code path touches (search for "not ok" in the "Search logs" box on the failed job's page).

dscho:
> I suspect that my `fixup!`s are incorrect

Well, one mistake that is quite obvious is that `!isdevice()` should be converted to `ondisk()` (note the missing `!` in the latter expression, which is what I missed 🤦 when I tried to do the rebase "on the side"). Let's see whether [inverting the logic as needed](https://github.com/git-for-windows/msys2-runtime/pull/84/commits/d6f33e5702de512f1ce46c6047837eeb4bed99dc) fixes the test failures.

lazka:
@dscho did you see the inline comment above? https://github.com/msys2/msys2-runtime/pull/254#pullrequestreview-2574383831

dscho:
> @dscho did you see the inline comment above? [#254 (review)](https://github.com/msys2/msys2-runtime/pull/254#pullrequestreview-2574383831)

Nope, I totally missed this. I am currently trying (since Friday) to convert a complex Azure Pipeline to a GitHub workflow (I really wish that GitHub invested _anything_ in GitHub Actions, including features to make debugging a lot less cumbersome), all while being sick, and I know that it's an idiotic idea of mine that I also do the MSYS2 runtime on the side. Sorry about that. Feel free to ignore me for now.

winsup/cygwin/path.cc
dscho: Actually, `(A || B) && A` reduces to `A` (where `A` is `!d.devn || d.devn == FH_FS` and `B` is `!d.devn || d.devn == FH_FIFO`). The `B` does not matter. It's a wasteful (and misleading) addition to the condition.

Therefore, the previous `!src_path.isdevice () && !src_path.is_fs_special ()` and the new `src_path.isondisk ()` are not only roughly identical, they are exactly identical (modulo performance if the compiler does not optimize the useless condition out), unless `dev_on_fs` does something that I missed.

lazka:
np, thanks for your work.

winsup/cygwin/path.cc
jeremyd2019: I don't know, I'm having a hard time reasoning about this, and I don't have a lot of attention to give this today, but the `dev_on_fs` vs `!dev_on_fs` is throwing me.

If it's `FH_FS` it doesn't matter.  The only other cases where it might is `devn == 0` or `FH_FIFO`, in which case `!dev_on_fs` would have had to have been true, now if it's not `FH_FS` `dev_on_fs` would have to be true instead.  I almost want to say `!src_path.isspecial()` might be what was intended here. But this is an odd edgecase (trying to figure out what to do if somebody tried to do a deepcopy symlink of a special file)

dscho: Oh darn it, forget my `(A || B) && A` reasoning: I missed that the first `A` starts with `!d.devn` and the second `A` starts with `!dev_on_fs`, i.e. the second `A` is actually a `C` and I was wrong...

dscho: Okay, let's try again:

```
(!d.devn || d.devn == FH_FS || !d.devn || d.devn == FH_FIFO) && (!dev_on_fs || d.devn == FH_FS)
```

Leaving out the duplicate `!d.devn`:

```
(!d.devn || d.devn == FH_FS || d.devn == FH_FIFO) && (!dev_on_fs || d.devn == FH_FS)
```

With `&&` binding stronger than `||` we can expand to:

```
!d.devn && !dev_on_fs || d.devn == FH_FS && !dev_on_fs || d.devn == FH_FIFO && !dev_on_fs || !d.devn && d.devn == FH_FS || d.devn == FH_FS && d.devn == FH_FS || d.devn == FH_FIFO && d.devn == FH_FS
```

Leaving out the impossible (like `!d.devn && d.devn  == FH_FS`):

```
!d.devn && !dev_on_fs || d.devn == FH_FS && !dev_on_fs || d.devn == FH_FIFO && !dev_on_fs || d.devn == FH_FS && d.devn == FH_FS
```

No, we can simplify `d.devn == FH_FS && d.devn == FH_FS` to `d.devn == FH_FS` and since that condition strictly encloses any condition `d.devn == FH_FS && ...` we can also drop the latter, i.e.

```
!d.devn && !dev_on_fs || d.devn == FH_FIFO && !dev_on_fs || d.devn == FH_FS
```

Since `dev_on_fs` can only be non-zero if `setfs(1)` is called, and in all three code-paths where that happens, the `devn` is parsed to a non-zero value first, `!d.devn` already implies `!dev_on_fs` and we can simplify the first term:

```
!d.devn || d.devn == FH_FIFO && !dev_on_fs || d.devn == FH_FS
```

Now, if `d.devn == FH_FIFO && !dev_on_fs` holds, then `d.devn != FH_FS` holds, too, (because it is implied) and therefore we can rewrite this condition as `d.devn == FH_FIFO && !(dev_on_fs || d.devn == FH_FS)` i.e. it is equivalent to `src_path.isfifo() && !src_path.isondisk()`

The `!d.devn || d.devn == FH_FS` condition can be rewritten as `!(d.devn && d.devn != FH_FS)`, which is equivalent to `!dev.not_device(FH_FS)`, i.e. `!src_path.isspecial()`.

So we can rewrite the original condition as `!src_path.isspecial() || (src_path.isfifo() && !src_path.isondisk())`.

However, that `src_path.isfifo() && !src_path.isondisk()` part looks suspicious. The code we're trying to adjust is the code path where the MSYS2 runtime is asked to make a symlink, and it makes a deep copy instead. Why would it make a deep copy of a FIFO, but only if it is not on disk (and would `CopyFile()` handle that at all, it would probably fail)? That part of the condition seems bogus.

Now, as the saying goes "If you want to know whether you truly understand something, try to explain it" and as the original code comment "MSYS copy file instead make symlink" explains precisely nothing at all, there is a very much non-zero chance that the original author did not understand the logic.

As a consequence, I maintain that the original condition was incorrect from the start. It would be not surprising at all, given that the entire idea of letting `symlink()` "succeed" by doing a deep copy is highly dubious to begin with, and I would love to have the time to push forward #114 further (which would require a deep dive into the question why Git's test suite fails with that).

In the meantime, and to unblock this PR, I would like to circle back to the question what condition _would_ be correct. We want to do a deep copy, and naturally we cannot copy certain entities (such as disk devices), and we would want to fail the operation in that case. I could imagine that the correct thing would be to drop the condition altogether: If a symlink should be emulated via a deep copy, and if that involves an entity that cannot be copied (because it is not on disk), then the operation should fail (instead of skipping the attempt to copy said entity).

Alternatively, I would suggest that we either go with the `!src_path.isspecial()` condition (which would align better with the original condition than `src_path.isondisk()`, and I just [verified](https://github.com/git-for-windows/msys2-runtime/actions/runs/12989859693?pr=84) that it at least does not break Git's test suite), or with `src_path_isondisk()`.

I would actually much prefer `src_path.isondisk()` because that would be a rare case of a readable and intuitively understandable construct in the code base.

@jeremyd2019 if you agree, I will reduce the `fixup!` commits into a single one so that I have a place where to jot down the reasoning in a commit message.

jeremyd2019: My understanding is that if that condition is false, it will instead fall back to making a `WSYM_sysfile` style symlink, not fail.  Windows apps (like Git for Windows) wouldn't know what to do with those, but they wouldn't know what to do with a Cygwin 'special' file either, so I expect that's ok.

I think `!src_path.isspecial()` because of that.

If there's a time and place to investigate/improve the deepcopy code (#114? or somewhere else), that looks like a good idea...  It only looks at the direct src_path to see if it's special, but then `recursiveCopy` just uses win32 functions, so it would have no idea if an on-disk special file was in there.  (also, I think it's using the ANSI Windows functions, plus is limited to `MAX_PATH`).

jeremyd2019: yeah, try
```console
$ mknod foo c 1 3
$ ln -s foo foolnk
$ ls -l foo*
crw-rw-rw- 1 x None 1, 3 Jan 27 10:43 foo
lrwxrwxrwx 1 x None    3 Jan 27 10:43 foolnk -> foo
```

see in Windows you've got `foo.lnk` for the device file (see the isdevice check at the beginning of the function, that comes from mknod not the former pathconv.isdevice()), and foolnk which is a sysfile-style symlink.

With the isondisk version currently present, as of [fa6f2c2](https://github.com/msys2/msys2-runtime/pull/254/commits/fa6f2c28b1b5f22ffd6e63902d4da62f51cde7b5), you get
```console
ls -l foo*
crw-rw-rw- 1 x None  1, 3 Jan 27 10:41 foo
-r--r--r-- 1 x None   126 Jan 27 10:41 foolnk

$ cat foolnk
L□F

:\1:3:21b6
:\1:3:21b6□□:\1:3:21b6
```

jeremyd2019: > ... `setfs(1)` is called, and in all three code-paths where that happens ...

BTW, this is back down to two code-paths after https://github.com/cygwin/cygwin/commit/5a58110d12842a477c4ff9f3f2df1c0175b134ce

dscho: > My understanding is that if that condition is false, it will instead fall back to making a `WSYM_sysfile` style symlink, not fail.

Ah, I missed that, thank you for pointing that out!

> try
> 
> ```
> $ mknod foo c 1 3
> $ ln -s foo foolnk
> $ ls -l foo*
> crw-rw-rw- 1 x None 1, 3 Jan 27 10:43 foo
> lrwxrwxrwx 1 x None    3 Jan 27 10:43 foolnk -> foo
> ```
> 
> see in Windows you've got `foo.lnk` for the device file (see the isdevice check at the beginning of the function, that comes from mknod not the former pathconv.isdevice()), and foolnk which is a sysfile-style symlink.

Excellent! I turned this into the following script:

```sh
#!/bin/sh

echo $PATH
for x in foo fifo
do
	test -e dir/$x || continue
	/c/WINDOWS/System32/WindowsPowerShell/v1.0/powershell \
		-command 'attrib.exe -R dir\'"$x"'.lnk' || exit 1
done &&
rm -rf dir &&
mkdir dir &&
cd dir &&
set -x &&

mknod foo c 1 3 &&
ln -s foo foolnk &&
ls -l foo* >out &&
cat out &&
grep '^l.*foolnk' out || {
	echo "Failed!" >&2
	exit 1
} &&

mkfifo fifo &&
ln -s fifo fifolnk &&
ls -l fifo* >out &&
cat out &&
grep '^l.*fifolnk' out || {
	echo "Failed!" >&2
	exit 1
}
```

This works with the current MSYS2 runtime (`uname -r` reports `3.5.5-1da33849.x86_64`).

And indeed, with the `isondisk()` method it does _fail_. And with `isspecial()` it _succeeds_. I will therefore go for the latter.

dscho: @jeremyd2019 thank you so much for your outstanding help. While looking at the code was far from pleasant, I enjoyed this collaboration more!

I reduced the fixups to a single one, as promised: a2cfafe8736436691b1fbe1bf2ea08812e78fa3e

dscho: And of course I managed to invert the logic again 🤦

dscho: Here is what I hope will be the tip of the final revision before we can go ahead and merge: 16f69cdec22aad0ddefaf3bb30dcebfe34175f62

dscho: And [here](https://github.com/git-for-windows/msys2-runtime/actions/runs/12997483342?pr=84) is the corresponding Git for Windows PR build that will also exercise Git's test suite.

jeremyd2019: *slight* editorial on commit message
> // otherwise fall back to creating a `.lnk` file

that fall back is the sysfile link type (`WSYM_sysfile`), not the `.lnk` link type (`WSYM_lnk`).

dscho: Fair! Feel free to make the change, otherwise I'll do it tomorrow.

jeremyd2019 (approved):

dscho:
⇄ https://github.com/msys2/MSYS2-packages/pull/5167

winsup/cygwin/path.cc
jeremyd2019: Another point in favor of `!isspecial()` being the proper condition:
https://github.com/msys2/msys2-runtime/blob/2c0e426da70f161af3776ffaa3559498b81c9df4/winsup/cygwin/path.cc#L1941-L1945

dscho: Excellent find!

jeremyd2019: > ```shell
> for x in foo fifo
> do
> 	test -e dir/$x || continue
> 	/c/WINDOWS/System32/WindowsPowerShell/v1.0/powershell \
> 		-command 'attrib.exe -R dir\'"$x"'.lnk' || exit 1
> done &&
> ```

I was about to debug why removing device files was failing, but I couldn't reproduce it with 3.5.6.  It looks like Corinna fixed that with the 002aad0410183153d8411405ed38c6c54c418ef5 change to `unlink` in syscalls.cc, so this workaround shouldn't be needed anymore.  Also, there was no need to call out to powershell, I was doing

```shell
for f in deepcopy/*; do attrib -r $f.lnk; rm $f.lnk; done
```

dscho: Right, I was testing with v3.5.5 (to verify that the expectations are correct), and at first tried a pure PowerShell solution to remove the files. It did not occur to me to skip PowerShell once it worked as desired...

PR #256, opened 2025-01-31
deepcopy symlink improvements

jeremyd2019:
Cleans up the deepcopy symlink code by:
* moving it into its own function
* making it use unicode internally, allowing for long paths and unicode characters.
* making it copy symlinks instead of their targets (wsl symlinks don't work until 19041, but everything else works all the way back to 8.1)
* adding more error checking

As part of this, I noticed this as well as a couple of other symlink handlers were doing the same thing to resolve the symlink targets.  I contributed 5a706ff0fceb83fd1fe7f072fc28a741fdde65f2 upstream, which this code can also use once it is rebased on top of that (presumably Cygwin 3.6)

### DIFF
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -2160,6 +2160,10 @@ symlink_deepcopy (const char *oldpath, path_conv &win32_newpath)
   tmp_pathbuf tp;
   path_conv win32_oldpath;
 
+  /* **BEGIN** replace this with
+     resolve_symlink_target (oldpath, win32_newpath. win32_oldpath);
+     when rebasing over 5a706ff0fceb83fd1fe7f072fc28a741fdde65f2
+     (probably Cygwin 3.6) */
   /* The symlink target is relative to the directory in which the
      symlink gets created, not relative to the cwd.  Therefore we
      have to mangle the path quite a bit before calling path_conv.*/
@@ -2174,6 +2178,7 @@ symlink_deepcopy (const char *oldpath, path_conv &win32_newpath)
 	      oldpath);
       win32_oldpath.check (absoldpath, PC_SYM_NOFOLLOW, stat_suffixes);
     }
+  /* **END** */
   if (win32_oldpath.error)
     {
       set_errno (win32_oldpath.error);
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -2073,6 +2073,79 @@ symlink_wsl (const char *oldpath, path_conv &win32_newpath)
 }
 
 int
+symlink_deepcopy (const char *oldpath, path_conv &win32_newpath)
+{
+  tmp_pathbuf tp;
+  path_conv src_path;
+
+  src_path.check (oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
+  if (src_path.error)
+    {
+      set_errno (src_path.error);
+      return -1;
+    }
+  if (src_path.isspecial ())
+    return -2;
+
+  /* MSYS copy file instead make symlink */
+  char * real_oldpath;
+  if (isabspath (oldpath))
+    strcpy (real_oldpath = tp.c_get (), oldpath);
+  else
+    /* Find the real source path, relative
+       to the directory of the destination */
+    {
+      /* Determine the character position of the last path component */
+      const char *newpath = win32_newpath.get_posix();
+      int pos = strlen (newpath);
+      while (--pos >= 0)
+	if (isdirsep (newpath[pos]))
+	  break;
+      /* Append the source path to the directory
+	 component of the destination */
+      if (pos+1+strlen(oldpath) >= MAX_PATH)
+	{
+	  set_errno(ENAMETOOLONG);
+	  return -1;
+	}
+      strcpy (real_oldpath = tp.c_get (), newpath);
+      strcpy (&real_oldpath[pos+1], oldpath);
+    }
+
+  /* As a MSYS limitation, the source path must exist. */
+  path_conv win32_oldpath;
+  win32_oldpath.check (real_oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
+  if (!win32_oldpath.exists ())
+    {
+      set_errno (ENOENT);
+      return -1;
+    }
+
+  char *w_newpath;
+  char *w_oldpath;
+  stpcpy (w_newpath = tp.c_get (), win32_newpath.get_win32());
+  stpcpy (w_oldpath = tp.c_get (), win32_oldpath.get_win32());
+  if (win32_oldpath.isdir())
+    {
+      char *origpath;
+      strcpy (origpath = tp.c_get (), w_oldpath);
+      return recursiveCopy (w_oldpath, w_newpath, origpath);
+    }
+  else
+    {
+      if (!CopyFile (w_oldpath, w_newpath, FALSE))
+	{
+	  __seterrno ();
+	  return -1;
+	}
+      else
+	{
+	  return 0;
+	}
+    }
+}
+
+int
 symlink_worker (const char *oldpath, path_conv &win32_newpath, bool isdevice)
 {
   int res = -1;
@@ -2140,6 +2213,13 @@ symlink_worker (const char *oldpath, path_conv &win32_newpath, bool isdevice)
 	case WSYM_nfs:
 	  res = symlink_nfs (oldpath, win32_newpath);
 	  __leave;
+	case WSYM_deepcopy:
+	  res = symlink_deepcopy (oldpath, win32_newpath);
+	  if (!res || res == -1)
+	    __leave;
+	  /* fall back to sysfile symlink type */
+	  wsym_type = WSYM_sysfile;
+	  break;
 	case WSYM_native:
 	case WSYM_nativestrict:
 	  res = symlink_native (oldpath, win32_newpath);
@@ -2308,77 +2388,6 @@ symlink_worker (const char *oldpath, path_conv &win32_newpath, bool isdevice)
 	}
       else /* wsym_type == WSYM_sysfile */
 	{
-          if (wsym_type == WSYM_deepcopy)
-	    {
-	      path_conv src_path;
-	      src_path.check (oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
-	      if (src_path.error)
-		{
-		  set_errno (src_path.error);
-		  __leave;
-		}
-	      if (!src_path.isspecial ())
-	        {
-		  /* MSYS copy file instead make symlink */
-
-		  char * real_oldpath;
-		  if (isabspath (oldpath))
-		    strcpy (real_oldpath = tp.c_get (), oldpath);
-		  else
-		    /* Find the real source path, relative
-		       to the directory of the destination */
-		    {
-		      /* Determine the character position of the last path component */
-		      const char *newpath = win32_newpath.get_posix();
-		      int pos = strlen (newpath);
-		      while (--pos >= 0)
-			if (isdirsep (newpath[pos]))
-			  break;
-		      /* Append the source path to the directory
-			 component of the destination */
-		      if (pos+1+strlen(oldpath) >= MAX_PATH)
-			{
-			  set_errno(ENAMETOOLONG);
-			  __leave;
-			}
-		      strcpy (real_oldpath = tp.c_get (), newpath);
-		      strcpy (&real_oldpath[pos+1], oldpath);
-		    }
-
-		  /* As a MSYS limitation, the source path must exist. */
-		  path_conv win32_oldpath;
-		  win32_oldpath.check (real_oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
-		  if (!win32_oldpath.exists ())
-		    {
-		      set_errno (ENOENT);
-		      __leave;
-		    }
-
-		  char *w_newpath;
-		  char *w_oldpath;
-		  stpcpy (w_newpath = tp.c_get (), win32_newpath.get_win32());
-		  stpcpy (w_oldpath = tp.c_get (), win32_oldpath.get_win32());
-		  if (win32_oldpath.isdir())
-		    {
-		      char *origpath;
-		      strcpy (origpath = tp.c_get (), w_oldpath);
-		      res = recursiveCopy (w_oldpath, w_newpath, origpath);
-		    }
-		  else
-		    {
-		      if (!CopyFile (w_oldpath, w_newpath, FALSE))
-			{
-			  __seterrno ();
-			}
-		      else
-			{
-			  res = 0;
-			}
-		    }
-		  __leave;
-		}
-	    }
-
 	  /* Default technique creating a symlink. */
 	  buf = tp.t_get ();
 	  cp = stpcpy (buf, SYMLINK_COOKIE);
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1735,6 +1735,7 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
   findfiles = FindNextFileW (dH, &dHfile);
   while (findfiles)
     {
+      bool isdirlink = false;
       /* Append the directory item filename to both source and destination */
       debug_printf ("dHfile(3): %W", dHfile.cFileName);
       src->Length = srcpos + sizeof (WCHAR);
@@ -1742,25 +1743,44 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
       RtlAppendUnicodeToString (src, dHfile.cFileName);
       RtlAppendUnicodeToString (dst, dHfile.cFileName);
       debug_printf ("%S -> %S", src, dst);
-      if (dHfile.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)
+      if ((dHfile.dwFileAttributes &
+	    (FILE_ATTRIBUTE_DIRECTORY|FILE_ATTRIBUTE_REPARSE_POINT)) ==
+	  (FILE_ATTRIBUTE_DIRECTORY|FILE_ATTRIBUTE_REPARSE_POINT))
+	{
+	  /* this sucks hard */
+	  path_conv pc (src, PC_SYM_NOFOLLOW|PC_SYM_NOFOLLOW_REP);
+	  isdirlink = pc.issymlink ();
+	}
+      if (isdirlink)
         {
-          /* Recurse into the child directory */
-          debug_printf ("%S <-> %W", src, origpath);
+	  /* CreateDirectoryEx seems to "copy" directory reparse points, which
+	     CopyFileEx can only do with a flag introduced in 19041. */
+	  if (!CreateDirectoryExW (src->Buffer, dst->Buffer, NULL))
+	    {
+	      debug_printf ("CreateDirectoryExW(%S, %S, 0) failed", src, dst);
+	      __seterrno ();
+	      goto done;
+	    }
+	}
+      else if (dHfile.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)
+	{
+	  /* Recurse into the child directory */
+	  debug_printf ("%S <-> %W", src, origpath);
 	  // avoids endless recursion
-          if (wcsncmp (src->Buffer, origpath, src->Length / sizeof (WCHAR)))
-            if (recursiveCopy (src, dst, origpath))
-              goto done;
-        }
+	  if (wcsncmp (src->Buffer, origpath, src->Length / sizeof (WCHAR)))
+	    if (recursiveCopy (src, dst, origpath))
+	      goto done;
+	}
       else
-        {
-          /* Just copy the file */
-          if (!CopyFileExW (src->Buffer, dst->Buffer, NULL, NULL, NULL,
+	{
+	  /* Just copy the file */
+	  if (!CopyFileExW (src->Buffer, dst->Buffer, NULL, NULL, NULL,
 			    COPY_FILE_COPY_SYMLINK))
-            {
-              __seterrno ();
-              goto done;
-            }
-        }
+	    {
+	      __seterrno ();
+	      goto done;
+	    }
+	}
       findfiles = FindNextFileW (dH, &dHfile);
     }
   if (GetLastError() != ERROR_NO_MORE_FILES)
@@ -2132,17 +2152,41 @@ symlink_deepcopy (const char *oldpath, path_conv &win32_newpath)
     }
   else
     {
-      if (!CopyFileExW (w_oldpath->Buffer, w_newpath->Buffer,
-			NULL, NULL, NULL, COPY_FILE_COPY_SYMLINK))
+      bool isdirlink = false;
+      if (win32_oldpath.issymlink () &&
+	  win32_oldpath.is_known_reparse_point ())
+	{
+	  /* Is there a better way to know this? */
+	  DWORD attr = getfileattr (win32_oldpath.get_win32 (),
+				    !!win32_oldpath.objcaseinsensitive ());
+	  if (attr == INVALID_FILE_ATTRIBUTES)
+	    {
+	      __seterrno ();
+	      return -1;
+	    }
+	  isdirlink = attr & FILE_ATTRIBUTE_DIRECTORY;
+	}
+      if (isdirlink)
+        {
+	  /* CreateDirectoryEx seems to "copy" directory reparse points, which
+	     CopyFileEx can only do with a flag introduced in 19041. */
+	  if (!CreateDirectoryExW (w_oldpath->Buffer, w_newpath->Buffer, NULL))
+	    {
+	      debug_printf ("CreateDirectoryExW(%S, %S, 0) failed", w_oldpath,
+			    w_newpath);
+	      __seterrno ();
+	      return -1;
+	    }
+	}
+      else if (!CopyFileExW (w_oldpath->Buffer, w_newpath->Buffer, NULL, NULL,
+			NULL, COPY_FILE_COPY_SYMLINK))
 	{
 	  __seterrno ();
 	  return -1;
 	}
-      else
-	{
-	  return 0;
-	}
     }
+
+  return 0;
 }
 
 int
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1697,7 +1697,6 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
 {
   WIN32_FIND_DATAW dHfile;
   HANDLE dH = INVALID_HANDLE_VALUE;
-  BOOL findfiles;
   int srcpos = src->Length;
   int dstpos = dst->Length;
   int res = -1;
@@ -1729,15 +1728,15 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
   dH = FindFirstFileExW (src->Buffer, FindExInfoBasic, &dHfile,
 			 FindExSearchNameMatch, NULL,
 			 FIND_FIRST_EX_LARGE_FETCH);
-  debug_printf ("dHfile(1): %W", dHfile.cFileName);
-  findfiles = FindNextFileW (dH, &dHfile);
-  debug_printf ("dHfile(2): %W", dHfile.cFileName);
-  findfiles = FindNextFileW (dH, &dHfile);
-  while (findfiles)
+  do
     {
       bool isdirlink = false;
+      debug_printf ("dHfile: %W", dHfile.cFileName);
+      if (dHfile.cFileName[0] == L'.' &&
+	  (!dHfile.cFileName[1] ||
+	   (dHfile.cFileName[1] == L'.' && !dHfile.cFileName[2])))
+	continue;
       /* Append the directory item filename to both source and destination */
-      debug_printf ("dHfile(3): %W", dHfile.cFileName);
       src->Length = srcpos + sizeof (WCHAR);
       dst->Length = dstpos + sizeof (WCHAR);
       RtlAppendUnicodeToString (src, dHfile.cFileName);
@@ -1781,8 +1780,9 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
 	      goto done;
 	    }
 	}
-      findfiles = FindNextFileW (dH, &dHfile);
     }
+  while (FindNextFileW (dH, &dHfile));
+
   if (GetLastError() != ERROR_NO_MORE_FILES)
     {
       __seterrno ();
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1693,69 +1693,75 @@ conv_path_list (const char *src, char *dst, size_t size,
   Create a deep copy of src as dst, while avoiding descending in origpath.
 */
 static int
-recursiveCopy (char * src, char * dst, const char * origpath)
+recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
 {
-  WIN32_FIND_DATA dHfile;
+  WIN32_FIND_DATAW dHfile;
   HANDLE dH = INVALID_HANDLE_VALUE;
   BOOL findfiles;
-  int srcpos = strlen (src);
-  int dstpos = strlen (dst);
+  int srcpos = src->Length;
+  int dstpos = dst->Length;
   int res = -1;
 
-  debug_printf("recursiveCopy (%s, %s)", src, dst);
+  debug_printf ("recursiveCopy (%S, %S)", src, dst);
 
   /* Create the destination directory */
-  if (!CreateDirectoryEx (src, dst, NULL))
+  if (!CreateDirectoryExW (src->Buffer, dst->Buffer, NULL))
     {
-      debug_printf("CreateDirectoryEx(%s, %s, 0) failed", src, dst);
+      debug_printf ("CreateDirectoryExW(%S, %S, 0) failed", src, dst);
       __seterrno ();
       goto done;
     }
   /* Descend into the source directory */
-  if (srcpos + 2 >= MAX_PATH || dstpos + 1 >= MAX_PATH)
+  if (src->Buffer[(src->Length - 1) / sizeof (WCHAR)] != L'\\')
     {
-      set_errno (ENAMETOOLONG);
-      goto done;
+      RtlAppendUnicodeToString (src, L"\\*");
     }
-  strcat (src, "\\*");
-  strcat (dst, "\\");
-  dH = FindFirstFile (src, &dHfile);
-  debug_printf("dHfile(1): %s", dHfile.cFileName);
-  findfiles = FindNextFile (dH, &dHfile);
-  debug_printf("dHfile(2): %s", dHfile.cFileName);
-  findfiles = FindNextFile (dH, &dHfile);
+  else
+    {
+      RtlAppendUnicodeToString (src, L"*");
+      srcpos -= sizeof (WCHAR);
+    }
+  if (dst->Buffer[(dst->Length - 1) / sizeof (WCHAR)] != L'\\')
+      RtlAppendUnicodeToString (dst, L"\\");
+  else
+      dstpos -= sizeof (WCHAR);
+
+  dH = FindFirstFileExW (src->Buffer, FindExInfoBasic, &dHfile,
+			 FindExSearchNameMatch, NULL,
+			 FIND_FIRST_EX_LARGE_FETCH);
+  debug_printf ("dHfile(1): %W", dHfile.cFileName);
+  findfiles = FindNextFileW (dH, &dHfile);
+  debug_printf ("dHfile(2): %W", dHfile.cFileName);
+  findfiles = FindNextFileW (dH, &dHfile);
   while (findfiles)
     {
       /* Append the directory item filename to both source and destination */
-      int filelen = strlen (dHfile.cFileName);
-      debug_printf("dHfile(3): %s", dHfile.cFileName);
-      if (srcpos + 1 + filelen >= MAX_PATH ||
-          dstpos + 1 + filelen >= MAX_PATH)
-        {
-          set_errno (ENAMETOOLONG);
-          goto done;
-        }
-      strcpy (&src[srcpos+1], dHfile.cFileName);
-      strcpy (&dst[dstpos+1], dHfile.cFileName);
-      debug_printf("%s -> %s", src, dst);
+      debug_printf ("dHfile(3): %W", dHfile.cFileName);
+      src->Length = srcpos + sizeof (WCHAR);
+      dst->Length = dstpos + sizeof (WCHAR);
+      RtlAppendUnicodeToString (src, dHfile.cFileName);
+      RtlAppendUnicodeToString (dst, dHfile.cFileName);
+      debug_printf ("%S -> %S", src, dst);
       if (dHfile.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)
         {
           /* Recurse into the child directory */
-          debug_printf("%s <-> %s", src, origpath);
-          if (strcmp (src, origpath)) // avoids endless recursion
+          debug_printf ("%S <-> %W", src, origpath);
+	  // avoids endless recursion
+          if (wcsncmp (src->Buffer, origpath, src->Length / sizeof (WCHAR)))
             if (recursiveCopy (src, dst, origpath))
               goto done;
         }
       else
         {
           /* Just copy the file */
-          if (!CopyFile (src, dst, FALSE))
+          if (!CopyFileExW (src->Buffer, dst->Buffer, NULL, NULL, NULL,
+			    COPY_FILE_COPY_SYMLINK))
             {
               __seterrno ();
               goto done;
             }
         }
-      findfiles = FindNextFile (dH, &dHfile);
+      findfiles = FindNextFileW (dH, &dHfile);
     }
   if (GetLastError() != ERROR_NO_MORE_FILES)
     {
@@ -2076,64 +2082,58 @@ int
 symlink_deepcopy (const char *oldpath, path_conv &win32_newpath)
 {
   tmp_pathbuf tp;
-  path_conv src_path;
+  path_conv win32_oldpath;
 
-  src_path.check (oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
-  if (src_path.error)
+  /* The symlink target is relative to the directory in which the
+     symlink gets created, not relative to the cwd.  Therefore we
+     have to mangle the path quite a bit before calling path_conv.*/
+  if (isabspath (oldpath))
+    win32_oldpath.check (oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
+  else
     {
-      set_errno (src_path.error);
+      size_t len = strrchr (win32_newpath.get_posix (), '/')
+		    - win32_newpath.get_posix () + 1;
+      char *absoldpath = tp.t_get ();
+      stpcpy (stpncpy (absoldpath, win32_newpath.get_posix (), len),
+	      oldpath);
+      win32_oldpath.check (absoldpath, PC_SYM_NOFOLLOW, stat_suffixes);
+    }
+  if (win32_oldpath.error)
+    {
+      set_errno (win32_oldpath.error);
       return -1;
     }
-  if (src_path.isspecial ())
+  if (win32_oldpath.isspecial ())
     return -2;
 
   /* MSYS copy file instead make symlink */
-  char * real_oldpath;
-  if (isabspath (oldpath))
-    strcpy (real_oldpath = tp.c_get (), oldpath);
-  else
-    /* Find the real source path, relative
-       to the directory of the destination */
-    {
-      /* Determine the character position of the last path component */
-      const char *newpath = win32_newpath.get_posix();
-      int pos = strlen (newpath);
-      while (--pos >= 0)
-	if (isdirsep (newpath[pos]))
-	  break;
-      /* Append the source path to the directory
-	 component of the destination */
-      if (pos+1+strlen(oldpath) >= MAX_PATH)
-	{
-	  set_errno(ENAMETOOLONG);
-	  return -1;
-	}
-      strcpy (real_oldpath = tp.c_get (), newpath);
-      strcpy (&real_oldpath[pos+1], oldpath);
-    }
-
   /* As a MSYS limitation, the source path must exist. */
-  path_conv win32_oldpath;
-  win32_oldpath.check (real_oldpath, PC_SYM_NOFOLLOW, stat_suffixes);
   if (!win32_oldpath.exists ())
     {
       set_errno (ENOENT);
       return -1;
     }
 
-  char *w_newpath;
-  char *w_oldpath;
-  stpcpy (w_newpath = tp.c_get (), win32_newpath.get_win32());
-  stpcpy (w_oldpath = tp.c_get (), win32_oldpath.get_win32());
-  if (win32_oldpath.isdir())
+  PUNICODE_STRING w_oldpath = win32_oldpath.get_nt_native_path ();
+  PUNICODE_STRING w_newpath = win32_newpath.get_nt_native_path ();
+  if (w_oldpath->Buffer[1] == L'?')
+    w_oldpath->Buffer[1] = L'\\';
+  if (w_newpath->Buffer[1] == L'?')
+    w_newpath->Buffer[1] = L'\\';
+  if (win32_oldpath.isdir ())
     {
-      char *origpath;
-      strcpy (origpath = tp.c_get (), w_oldpath);
-      return recursiveCopy (w_oldpath, w_newpath, origpath);
+      PWCHAR origpath = win32_oldpath.get_wide_win32_path (tp.w_get ());
+      /* we need a larger UNICODE_STRING MaximumLength than
+	 get_nt_native_path allocates for the recursive copy */
+      UNICODE_STRING u_oldpath, u_newpath;
+      RtlCopyUnicodeString (tp.u_get (&u_oldpath), w_oldpath);
+      RtlCopyUnicodeString (tp.u_get (&u_newpath), w_newpath);
+      return recursiveCopy (&u_oldpath, &u_newpath, origpath);
     }
   else
     {
-      if (!CopyFile (w_oldpath, w_newpath, FALSE))
+      if (!CopyFileExW (w_oldpath->Buffer, w_newpath->Buffer,
+			NULL, NULL, NULL, COPY_FILE_COPY_SYMLINK))
 	{
 	  __seterrno ();
 	  return -1;
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1706,7 +1706,8 @@ recursiveCopyCheckSymlink(PUNICODE_STRING src, bool& isdirlink)
   Create a deep copy of src as dst, while avoiding descending in origpath.
 */
 static int
-recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath, PWIN32_FIND_DATAW dHfile = NULL)
+recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, USHORT origsrclen,
+	       USHORT origdstlen, PWIN32_FIND_DATAW dHfile = NULL)
 {
   HANDLE dH = INVALID_HANDLE_VALUE;
   NTSTATUS status;
@@ -1812,11 +1813,15 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath, PWIN32
       else if (dHfile->dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)
 	{
 	  /* Recurse into the child directory */
-	  debug_printf ("%S <-> %W", src, origpath);
-	  // avoids endless recursion
-	  if (wcsncmp (src->Buffer, origpath, src->Length / sizeof (WCHAR)))
-	    if (recursiveCopy (src, dst, origpath, dHfile))
+	  /* avoids endless recursion */
+	  if (src->Length <= origsrclen ||
+	      !wcsncmp (src->Buffer, dst->Buffer, origdstlen / sizeof (WCHAR)))
+	    {
+	      set_errno (ELOOP);
 	      goto done;
+	    }
+	  if (recursiveCopy (src, dst, origsrclen, origdstlen, dHfile))
+	    goto done;
 	}
       else
 	{
@@ -2193,13 +2198,13 @@ symlink_deepcopy (const char *oldpath, path_conv &win32_newpath)
     w_newpath->Buffer[1] = L'\\';
   if (win32_oldpath.isdir ())
     {
-      PWCHAR origpath = win32_oldpath.get_wide_win32_path (tp.w_get ());
       /* we need a larger UNICODE_STRING MaximumLength than
 	 get_nt_native_path allocates for the recursive copy */
       UNICODE_STRING u_oldpath, u_newpath;
       RtlCopyUnicodeString (tp.u_get (&u_oldpath), w_oldpath);
       RtlCopyUnicodeString (tp.u_get (&u_newpath), w_newpath);
-      return recursiveCopy (&u_oldpath, &u_newpath, origpath);
+      return recursiveCopy (&u_oldpath, &u_newpath,
+			    u_oldpath.Length, u_newpath.Length);
     }
   else
     {
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1697,6 +1697,7 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
 {
   WIN32_FIND_DATAW dHfile;
   HANDLE dH = INVALID_HANDLE_VALUE;
+  NTSTATUS status;
   int srcpos = src->Length;
   int dstpos = dst->Length;
   int res = -1;
@@ -1713,21 +1714,37 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
   /* Descend into the source directory */
   if (src->Buffer[(src->Length - 1) / sizeof (WCHAR)] != L'\\')
     {
-      RtlAppendUnicodeToString (src, L"\\*");
+      status = RtlAppendUnicodeToString (src, L"\\*");
     }
   else
     {
-      RtlAppendUnicodeToString (src, L"*");
+      status = RtlAppendUnicodeToString (src, L"*");
       srcpos -= sizeof (WCHAR);
     }
+  if (!NT_SUCCESS (status))
+    {
+      __seterrno_from_nt_status (status);
+      goto done;
+    }
   if (dst->Buffer[(dst->Length - 1) / sizeof (WCHAR)] != L'\\')
-      RtlAppendUnicodeToString (dst, L"\\");
+      status = RtlAppendUnicodeToString (dst, L"\\");
   else
       dstpos -= sizeof (WCHAR);
+  if (!NT_SUCCESS (status))
+    {
+      __seterrno_from_nt_status (status);
+      goto done;
+    }
 
   dH = FindFirstFileExW (src->Buffer, FindExInfoBasic, &dHfile,
 			 FindExSearchNameMatch, NULL,
 			 FIND_FIRST_EX_LARGE_FETCH);
+  if (dH == INVALID_HANDLE_VALUE)
+    {
+      __seterrno ();
+      goto done;
+    }
+
   do
     {
       bool isdirlink = false;
@@ -1739,8 +1756,18 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
       /* Append the directory item filename to both source and destination */
       src->Length = srcpos + sizeof (WCHAR);
       dst->Length = dstpos + sizeof (WCHAR);
-      RtlAppendUnicodeToString (src, dHfile.cFileName);
-      RtlAppendUnicodeToString (dst, dHfile.cFileName);
+      status = RtlAppendUnicodeToString (src, dHfile.cFileName);
+      if (!NT_SUCCESS (status))
+	{
+	  __seterrno_from_nt_status (status);
+	  goto done;
+	}
+      status = RtlAppendUnicodeToString (dst, dHfile.cFileName);
+      if (!NT_SUCCESS (status))
+	{
+	  __seterrno_from_nt_status (status);
+	  goto done;
+	}
       debug_printf ("%S -> %S", src, dst);
       if ((dHfile.dwFileAttributes &
 	    (FILE_ATTRIBUTE_DIRECTORY|FILE_ATTRIBUTE_REPARSE_POINT)) ==
@@ -1748,6 +1775,11 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
 	{
 	  /* this sucks hard */
 	  path_conv pc (src, PC_SYM_NOFOLLOW|PC_SYM_NOFOLLOW_REP);
+	  if (pc.error)
+	    {
+	      set_errno (pc.error);
+	      goto done;
+	    }
 	  isdirlink = pc.issymlink ();
 	}
       if (isdirlink)
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1689,18 +1689,37 @@ conv_path_list (const char *src, char *dst, size_t size,
 
 /********************** Symbolic Link Support **************************/
 
+static int
+recursiveCopyCheckSymlink(PUNICODE_STRING src, bool& isdirlink)
+{
+  path_conv pc (src, PC_SYM_NOFOLLOW|PC_SYM_NOFOLLOW_REP);
+  if (pc.error)
+    {
+      set_errno (pc.error);
+      return -1;
+    }
+  isdirlink = pc.issymlink ();
+  return 0;
+}
+
 /*
   Create a deep copy of src as dst, while avoiding descending in origpath.
 */
 static int
-recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
+recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath, PWIN32_FIND_DATAW dHfile = NULL)
 {
-  WIN32_FIND_DATAW dHfile;
   HANDLE dH = INVALID_HANDLE_VALUE;
   NTSTATUS status;
   int srcpos = src->Length;
   int dstpos = dst->Length;
   int res = -1;
+  bool freedHfile = false;
+
+  if (!dHfile)
+    {
+      dHfile = (PWIN32_FIND_DATAW) cmalloc_abort (HEAP_STR, sizeof (*dHfile));
+      freedHfile = true;
+    }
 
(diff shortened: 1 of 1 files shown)


jeremyd2019:
Well, I tried these changes on my git-for-windows fork and the git test suite all passes, plus the tests I contrived are working.  I think this is at least as good as what was there before... With one *interesting* side effect:
```shell
mkdir dir
cd dir
ln -s ../dir .
```
before it failed with:
```
ln: failed to create symbolic link './dir' -> '../dir': File name too long
$ find dir | wc
     55      55    6160
```

now it fails with:
```
ln: failed to create symbolic link './dir': Bad address
$ find dir | wc
   1894    1894 7178260
```

Removing the `MAX_PATH` limit means it now exhausts stack recursing before it hits the new limit (of 32k iirc). 🤦

jeremyd2019:
oh, I need to dig up a Windows 8.1 VM and test this there, so far I've only run it on windows 11 23H2 (and apparently server 2022 on the gfw test suite)

jeremyd2019:
```
$ find dir | wc
   4155    4155 34536360
```

jeremyd2019:
Does anyone know/remember what the origpath attempt to avoid endless recursion was supposed to be protecting against?  It seems to me that it would never be triggered, and it would perhaps make more sense to check that src doesn't recurse into the original dst instead (maybe that's what it was supposed to be doing and using `w_oldpath` was a bug?)

jeremyd2019:
```console
$ ln -s ../dir .
ln: failed to create symbolic link './dir': Permission denied
$ find dir | wc
find: ‘dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir/dir’: No such file or directory
   8180    8180 133841160
```

Yay?

jeremyd2019:
```console
$ ln -s ../dir .
ln: failed to create symbolic link './dir': Too many levels of symbolic links
```

which is much better, but it was good for exercising the deep recursion anyway.

jeremyd2019:
Changing the fallback of deepcopy from sysfile to default makes a git test fail:
```
2025-02-01T00:52:31.0466688Z ++ mkdir dir
2025-02-01T00:52:31.0466758Z ++ ln -s /dev/null dir/subdir
2025-02-01T00:52:31.0466813Z ++ git add .
2025-02-01T00:52:31.0466913Z error: open("dir/subdir"): Function not implemented
```

In that case, a link to a special file is a wsl symlink rather than a sysfile... hmm

jeremyd2019:
wsl symlinks also don't play well with CopyFileEx on windows 8.1, so I've switched the fallback for deepcopy back to sysfile.

dscho (approved):
Looks good to me!

dscho:
Given that MSYS2 will most likely drop support for Windows versions that do not support symlinks (= Windows 8.1) and for Windows versions that require elevation to create symlinks (= Windows 10 prior to Build 14972), I consider the deepcopy mode of less interest than if it had to be maintained for the foreseeable future.

Luckily, we will be able to drop the deepcopy code altogether within the next couple of years. Having said that, it'll be nice to have it in a better shape than it currently is, especially for users stuck with older Windows versions.

jeremyd2019:
> FWIW @jeremyd2019 I really like the refactoring you did, moving the deepcopy code into its own function, I'd just suggest to do it as a preparatory, separate commit instead of squashing it in.

Please explain.  You would like https://github.com/msys2/msys2-runtime/pull/256/commits/6cc882399ad1da6f6803f295b296bd4a29ef6ad8 to be done before the other changes in this PR?  Or you think this shouldn't be a `fixup!`?

(I'm planning to rebase the commits in this PR anyway (I was thinking about using `amend!`) before switching out of draft mode)

jeremyd2019:
> Windows versions that require elevation to create symlinks (= Windows 10 prior to Build 14972)

I'm on Windows 11 22631, and it still requires elevation to create symlinks.  I haven't turned on developer mode though (I thought it had to do with UWP development, not sure of all the implications).

> Windows versions that do not support symlinks (= Windows 8.1) 

The docs say CreateSymbolicLinkW is available on Vista or later, and they seemed to work on 8.1 when I was testing this PR

jeremyd2019:
I tested the current state of this PR on various Windows versions I expected to be relevant, here's the result:
| Link type | 9600 | 10240 | 14393 | 16299 | 19041 | 22000 | 22631 |
| --------- | ---- | ----- | ----- | ----- | ----- | ----- | ----- |
| wsl | :x: | :x: | :x: | :x: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| junction | :x: | :x: | :x: | :x: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| symlinkd | :x: | :x: | :x: | :x: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| symlink | :white_check_mark: | :white_check_mark:  | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |

While doing this, I accidentally started testing the existing code instead, and saw some odd behavior that suggests I can actually improve on this on older Windows versions...

jeremyd2019:
After bbe61013, my test table looks like:
| Link type | 9600 | 10240 | 14393 | 16299 | 19041 | 22000 | 22631 |
| --------- | ---- | ----- | ----- | ----- | ----- | ----- | ----- |
| wsl | :x: | :x: | :x: | :x: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| junction | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| symlinkd | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| symlink | :white_check_mark: | :white_check_mark:  | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |

jeremyd2019:
Please advise on rebasing/squashing.

winsup/cygwin/path.cc
```diff
@@ -1748,47 +1767,44 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, PCWSTR origpath)
...
 	}
-      if ((dHfile.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) && !isdirlink)
+      if ((dHfile->dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) && !isdirlink)
 	{
 	  /* Recurse into the child directory */
 	  debug_printf ("%S <-> %W", src, origpath);
 	  // avoids endless recursion
 	  if (wcsncmp (src->Buffer, origpath, src->Length / sizeof (WCHAR)))
-	    if (recursiveCopy (src, dst, origpath))
+	    if (recursiveCopy (src, dst, origpath, dHfile))
```
dscho: I do not believe that this can work correctly: `dHfile` still holds the state for the directory enumeration, but in the nested call is used to store the state for _another_ directory enumeration. Most likely the outer loop's next `FindNextFileW()` call (25 lines below) will return `FALSE` because of the _inner loop_'s directory enumeration state signaling that it is done.

An alternative to spending 592 bytes per recursion would be to build up a stack of `src`/`dst`/`origpath` triplets, initializing it with the top-most directories' paths, and then running the (obviously to-be-renamed) `recursiveCopy()` in a `while` loop, `pop()`ing one entry at a time (and `push()`ing to it instead of recursing).

This stack would also avoid cluttering up the stack space, as it would internally use `malloc()`.

dscho:
> > FWIW @jeremyd2019 I really like the refactoring you did, moving the deepcopy code into its own function, I'd just suggest to do it as a preparatory, separate commit instead of squashing it in.
> 
> Please explain. You would like [6cc8823](https://github.com/msys2/msys2-runtime/commit/6cc882399ad1da6f6803f295b296bd4a29ef6ad8) to be done before the other changes in this PR? Or you think this shouldn't be a `fixup!`?

Right, my answer was quite confusing. Yes on both. I would like the refactoring to be done before the other changes (because the refactoring could even be applied independently), and due to the way the `msys2-runtime` repository works, it should still be marked as a `fixup!`.

But having had a look at the current state of the PullRequest, I have no qualm about leaving the refactoring in the middle, either.

> > Windows versions that require elevation to create symlinks (= Windows 10 prior to Build 14972)
> 
> I'm on Windows 11 22631, and it still requires elevation to create symlinks. I haven't turned on developer mode though (I thought it had to do with UWP development, not sure of all the implications).

Indeed, even the latest Windows version unfortunately requires Developer Mode to allow creating symlinks in an unelevated session.

> > Windows versions that do not support symlinks (= Windows 8.1)
> 
> The docs say CreateSymbolicLinkW is available on Vista or later, and they seemed to work on 8.1 when I was testing this PR

I should have been more clear: I consider symlinks that can be deleted without elevation but that cannot be recreated without elevation not to be symlinks at all but a horrible design bug. As such, I consider any Windows version prior to Build 14972 not _actually_ to have support for symlinks.

> After [bbe6101](https://github.com/msys2/msys2-runtime/commit/bbe61013eb97f6220c813e7ec7bbbccbbbc1f7ef), my test table looks like:
> 
> | Link type | 9600 | 10240 | 14393 | 16299 | 19041 | 22000 | 22631
> | - | - | - | - | - | - | - | - |
> | wsl | ❌ | ❌ | ❌ | ❌ | ✅ | ✅ | ✅ |
> | junction | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
> | symlinkd | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
> | symlink | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |

Very nice!!!

I just had a look through the commits and have only two suggestions. I'll finalize the review properly just after posting this here comment.

dscho (changes requested):
I only have one more wish: Add yet another `fixup!` commit, this one generated using `git commit --fixup reword:44054a96275c0f058069e6bb39b3b859157b85bd -s` to amend the commit message (and to take co-ownership because of your extensive improvements).

winsup/cygwin/path.cc
jeremyd2019: I don't think it's a problem (obviously, since I wrote it 😉), `WIN32_FIND_DATAW` is documented as purely an `[out]` parameter, not an `[in, out]`.  Any state should be held by the `HANDLE` (`dH`), of which there is one per stack frame.

dscho: Oh, wow, TIL! It [indeed is an `[out]` parameter](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findnextfilew#syntax)... You're right, and I retract my erroneous objection.

jeremyd2019:
OK, I've rebased now, moved the factoring out to a function first, squashed the COPY_FILE_DIRECTORY vs CreateDirectoryEx change into the commit that added handling of directory links, and added an `amend!` commit that adds Co-authored-by (and a comment around the code to be replaced by a new function).  all the intermediate commits at least build successfully in CI.  Is there anything else you think should be added to the commit message in the `amend!` commit?

BTW, doing an interactive rebase on this was a real pain, because `git rebase -i` only shows the first line of the commit message, and that's the same for all of these commits.  I had to refer to this PR page commits tab with the messages expanded to match up each commit hash to what it actually for.

dscho:
> BTW, doing an interactive rebase on this was a real pain, because `git rebase -i` only shows the first line of the commit message, and that's the same for all of these commits.

☹️ Yes, `rebase -i` can be a real pain to use.

FWIW there is [`rebase.instructionFormat`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt-rebaseinstructionFormat), but it lacks any convenient way to display the body (or an excerpt thereof), and it does not handle `%N` for notes. The only way to make that somewhat work would be to tag the individual commits with names like `refactor-recursive-copy` or some such.

dscho (approved):

PR #262, opened 2025-03-03
Fix console mode restoring

dscho:
Under certain circumstances, when a non-MSYS process is spawned from an MSYS process that is spawned from a non-MSYS process that is spawned from an MSYS process, the console state can get corrupted by the MSYS2 runtime.

This is the case, for example, [when spawning a background process that outputs to `stdout` from a Git hook](https://github.com/microsoft/git/issues/730).

This behavior is inherited from the Cygwin runtime which allows the same symptom to be reproduced as long as one is careful enough to force `git.exe` to use Cygwin's Bash to run its hooks.

The Cygwin core developer who is the only person to understand Cygwin's pseudo console handling was kind enough to fix this problem, and I backported it to the MSYS2 runtime v3.5.7.

This is a companion of https://github.com/git-for-windows/msys2-runtime/pull/87

### DIFF
--- a/winsup/cygwin/fhandler/console.cc
+++ b/winsup/cygwin/fhandler/console.cc
@@ -916,8 +916,7 @@ fhandler_console::cleanup_for_non_cygwin_app (handle_set_t *p)
   /* Cleaning-up console mode for non-cygwin app. */
   /* conmode can be tty::restore when non-cygwin app is
      exec'ed from login shell. */
-  tty::cons_mode conmode =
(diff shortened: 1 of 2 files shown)


dscho:
That was quick ;-)

dscho:
@lazka do you want me to take care of msys2/MSYS2-packages? (We should probably think about some sort of automation in the long run, so that this does not have to be done manually all the time.)

lazka:
sure, will do.

dscho:
> sure, will do.

Thank you!

PR #269, opened 2025-03-21
Cygwin: console: tty::restore really restores the previous mode

### LINKED ISSUE #268: Python Textual app doesn't work properly on MSYS2 run through Command Prompt after 'msys2-runtime' update to >3.5.7-2-x86_64
My team have been running a Python Textual app in MSYS2's UCRT64 environment through a Command Prompt terminal. We had to use the Command Prompt terminal instead of the default `mintty` because the Textual app could not be clicked when run in a `mintty` termnial. 

However, the recent `msys2-runtime` package update to >3.5.7-2-x86_64 has caused the app to glitch out when switching to a new screen. Instead of displaying a screen, it prints out characters according to mouse movement within the console window area.

Understanding what changed between version 3.5.7-2-x86_64 and 3.5.7-4-x86_64 and seeing what can be done to resolve it would be very appreciated.

(Additional context below)

**The Problem:**
- When running our app using MSYS2's default `mintty` terminal, we can't click on any of the buttons in the console.
- This behaviour is present across the wide range of Windows 10 versions we work with (Windows 10 Server 2016 & 2019, Windows 10 Professional & Home).
- This is unique to Windows 10, as users on the MSYS2 Discord channel have not been able to replicate the issue on Windows 11.
- This problem is not present in Cygwin's `mintty` terminal on these same Windows 10 versions; the Textual app's buttons all work as intended.
- These are externally managed PCs, so updating them is out of the question.

**What We've Tried So Far:**
- This problem does not occur when MSYS2 is run via Windows Terminal.
  - It is not supported for the Windows 10 versions present on the PCs we manage, so it is not a viable solution.
- We found a workaround where we run an MSYS2 session in a Command Prompt console.
  - This involved creating a Command Prompt shortcut and pointing to MSYS2 in the "Target" field as follows: `cmd.exe /k "C:\msys64\msys2_shell.cmd -defterm -no-start -ucrt64 -shell bash"`.
  - The Textual app runs in that case; the buttons are clickable, although the UI occasionally displays funny characters.
- Updating `msys2-runtime` to >3.5.7-2-x86_64 causes this Command Prompt console workaround to fail.

tieneupin:
Previously, tty::restore sets the console mode to a predetermined console mode that is widely common for many non-cygwin console apps. So, if a non-cygwin app that is started from cygwin process changes the console mode and executes cygwin sub-process, the console mode is changed to the predetermined mode rather than being restored the original mode that is set by the non-cygwin app.
With this patch, the console mode is stored when a cygwin process is started from non-cygwin app, then tty::restore restores the previous console mode that is used by the previous non-cygwin app when the cygwin app exits.

Addresses: https://github.com/msys2/msys2-runtime/issues/268
Fixes: 3312f2d21f13 ("Cygwin: console: Redesign mode set strategy on close().")
Reported-by: Eu Pin Tien, Jeremy Drake <cygwin@jdrake.com>

### DIFF
--- a/winsup/cygwin/fhandler/console.cc
+++ b/winsup/cygwin/fhandler/console.cc
@@ -804,6 +804,9 @@ fhandler_console::rabuflen ()
   return con_ra.rabuflen;
 }
 
+static DWORD prev_input_mode_backup;
+static DWORD prev_output_mode_backup;
+
 /* The function set_{in,out}put_mode() should be static so that they
    can be called even after the fhandler_console instance is deleted. */
 void
@@ -818,11 +821,11 @@ fhandler_console::set_input_mode (tty::cons_mode m, const termios *t,
   GetConsoleMode (p->input_handle, &oflags);
   DWORD flags = oflags
     & (ENABLE_EXTENDED_FLAGS | ENABLE_INSERT_MODE | ENABLE_QUICK_EDIT_MODE);
-  con.curr_input_mode = m;
   switch (m)
     {
     case tty::restore:
-      flags |= ENABLE_ECHO_INPUT | ENABLE_LINE_INPUT | ENABLE_PROCESSED_INPUT;
+      flags = con.prev_input_mode;
+      con.prev_input_mode = prev_input_mode_backup;
       break;
     case tty::cygwin:
       flags |= ENABLE_WINDOW_INPUT;
@@ -846,6 +849,12 @@ fhandler_console::set_input_mode (tty::cons_mode m, const termios *t,
 	flags |= ENABLE_PROCESSED_INPUT;
       break;
     }
+  if (con.curr_input_mode != tty::cygwin && m == tty::cygwin)
+    {
+      prev_input_mode_backup = con.prev_input_mode;
+      con.prev_input_mode = oflags;
+    }
+  con.curr_input_mode = m;
   SetConsoleMode (p->input_handle, flags);
   if (!(oflags & ENABLE_VIRTUAL_TERMINAL_INPUT)
       && (flags & ENABLE_VIRTUAL_TERMINAL_INPUT)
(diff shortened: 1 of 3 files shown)


dscho:
@tieneupin excellent. Have you verified that it fixes your problem?

tieneupin:
@dscho , YES. 🎉 The fix worked. The rsync subprocess ran smoothly through to completion and directed me to the next screen. 🙌

dscho:
> @dscho , YES. 🎉 The fix worked. The rsync subprocess ran smoothly through to completion and directed me to the next screen. 🙌

Excellent!

There is just one more complication, as this here PR crossed paths with the upgrade to Cygwin v3.6.0. Let me rebase the fix.

dscho:
@tieneupin I [changed](https://github.com/msys2/msys2-runtime/pull/269#event-16931228827) the base and [rebased](https://github.com/msys2/msys2-runtime/compare/a0fa7d3c0b571d509180f510d2941a02e510b8cd..c1770e171505eb674626c2b7abf3403e6b4b7b79) the patch. The range-diff looks like this:

* 1:  a0fa7d3c0b ! 1:  c1770e1715 Cygwin: console: tty::restore really restores the previous mode

   ``````diff
   @@ Commit message
        Fixes: 3312f2d21f13 ("Cygwin: console: Redesign mode set strategy on close().")
        Reported-by: Eu Pin Tien, Jeremy Drake <cygwin@jdrake.com>
        Signed-off-by: Takashi Yano <takashi.yano@nifty.ne.jp>
   +    Signed-off-by: Eu-Pin Tien <eu-pin.tien@diamond.ac.uk>
   +    Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>
    
     ## winsup/cygwin/fhandler/console.cc ##
    @@ winsup/cygwin/fhandler/console.cc: fhandler_console::rabuflen ()
   @@ winsup/cygwin/fhandler/console.cc: fhandler_console::cleanup_for_non_cygwin_app
       set_output_mode (conmode, ti, p);
       set_input_mode (conmode, ti, p);
       set_disable_master_thread (con.owner == GetCurrentProcessId ());
   -@@ winsup/cygwin/fhandler/console.cc: fhandler_console::close ()
   +@@ winsup/cygwin/fhandler/console.cc: fhandler_console::close (int flag)
       if (shared_console_info[unit] && myself->ppid == 1
           && (dev_t) myself->ctty == get_device ())
         {
   ``````

dscho:
→ https://github.com/msys2/MSYS2-packages/pull/5269

(sorry, this took a bit longer than planned.)

PR #271, opened 2025-03-27
Cygwin: pipe: Fix hang due to inadvertent 0-length raw_write()

### LINKED ISSUE #270: GIMP build hangs forever with 3.6.0
https://github.com/msys2/MINGW-packages/actions/runs/14042781889/job/39316680160

From what I see if python calls into cygwin binaries things hang forever, or something like that. It happens consistently.

I can reproduce locally. Downgrading to 3.5.7 fixes things.

dscho:
It is possible for `NtQueryInformationFile()` to report a 0-length `InboundQuota` when obtaining `FilePipeLocalInformation`. This seems to be the case e.g. when a pipe was created on the other side and its quota information is not available on the client side.

This can lead to a situation where the `avail` variable is set to 0, and since that is used to cap the number of bytes to send, a 0-length write. Which hangs forever.

This was observed in the MSYS2 project when building GIMP, and reduced to a simple test case where a MINGW `ninja.exe` tries to call an MSYS `bison.exe` and the error message (saying that `bison` wants to have some input) is not even shown.

Since the minimal pipe buffer size is 4k, let's ensure that it is at least that, even when `InboundQuota` reports 0.

This fixes https://github.com/msys2/msys2-runtime/issues/270

I am also [upstreaming this patch](https://inbox.sourceware.org/cygwin-patches/40a66dcf272eff407794bb94616d5b8ba833fafa.1743076896.git.johannes.schindelin@gmx.de/) to the Cygwin project as we speak.

### DIFF
--- a/winsup/cygwin/fhandler/pipe.cc
+++ b/winsup/cygwin/fhandler/pipe.cc
@@ -1158,7 +1158,7 @@ fhandler_pipe::set_pipe_buf_size ()
   status = NtQueryInformationFile (get_handle (), &io, &fpli, sizeof fpli,
 				   FilePipeLocalInformation);
   if (NT_SUCCESS (status))
-    pipe_buf_size = fpli.InboundQuota;
+    pipe_buf_size = fpli.InboundQuota < PIPE_BUF ? PIPE_BUF : fpli.InboundQuota;
 }
 
 int


tyan0:
@dscho, Thanks for tracking down the problem! I'll check why the InboundQuota returns 0. Please wait a while.

dscho:
I haven't seen any more traction on the Cygwin-patches mailing list or on IRC, and I agree with @lazka that we should apply this until the time when upstream Cygwin fixes this differently (if that time comes 😉; my fix might make it in as-is, after all).

oold:
> I haven't seen any more traction on the Cygwin-patches mailing list or on IRC, and I agree with @lazka that we should apply this until the time when upstream Cygwin fixes this differently (if that time comes 😉; my fix might make it in as-is, after all).

Great. I think we're dealing with exactly that problem in Cygwin right now. Trying to get our Cygwin-hosted toolchain to run with Windows-native CMake and Ninja. GCC hangs during CMake's "detecting compiler ABI" and, for some reason, when compiling ASM files, but not C/C++ files.

@dscho Any idea why they're not attempting to merge your fix?

lazka:
This specific issue was fixed in upstream cygwin long ago: https://github.com/msys2/msys2-runtime/commit/d750786e7de013b58e2968eeb6a7fd59dcc535c9

oold:
Oh, thank you very much. I didn't know. We're still on v3.6.0. I'll try to upgrade then. Have a nice weekend!

PR #278, opened 2025-04-10
fix infinite recursion check to only match subdirectories

### LINKED ISSUE #277: Symbolic link creation failure in deepcopy mode
I found the problem in the latest msys2 runtime which symbolic link creation fails with "Too many levels of symbolic links" error.

The problem occurs under the following conditions:
* Symlink mode is deepcopy (`MSYS=winsymlinks:deepcopy`)
* Create a symbolic link to the existing directory which contains sub-directory.
* The link target name is substring of source name.

For example, the following operation fails with error.
```
$ export MSYS=   # set deepcopy mode
$ mkdir -p abcd/a
$ ln -s abcd a
ln: failed to create symbolic link 'a': Too many levels of symbolic links
$ ln -s abcd ab
ln: failed to create symbolic link 'ab': Too many levels of symbolic links
$ ln -s abcd abc
ln: failed to create symbolic link 'abc': Too many levels of symbolic links
```
The following cases are OK.
```
$ ln -s abcd aa
$ ln -s abcd abcde
```

I confirmed this problem occurs in msys2-runtime `3.5.7-4`. `3.5.3-4` doesn't have the problem.

jeremyd2019:
Adjust the infinite recursion check to make sure the source ends with a dir sep, so it doesn't think abcd is a subdirectory of a.

Fixes #277

### DIFF
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -1855,7 +1855,9 @@ recursiveCopy (PUNICODE_STRING src, PUNICODE_STRING dst, USHORT origsrclen,
 	  /* Recurse into the child directory */
 	  /* avoids endless recursion */
 	  if (src->Length <= origsrclen ||
-	      !wcsncmp (src->Buffer, dst->Buffer, origdstlen / sizeof (WCHAR)))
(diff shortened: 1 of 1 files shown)


dscho (approved):
Looks obviously good!

PR #279, opened 2025-04-10
Cygwin: console: Fix the console states after the console is closed

jeremyd2019:
Due to a bug introduced by the commit 3312f2d21f13, when the parent process exits before the child process exits, disable_master_thread is wrongly set to true, that disables special key handling such as Ctrl-C. With this patch, the disable_master_thread is set to true if any of the following conditions is met.
  - The parent process is not a cygwin process.
  - The master process already died.
  - The current process is the master process. Otherwise, disable_master_thread remains false to keep special key handling enabled.

Addresses: https://cygwin.com/pipermail/cygwin/2025-April/257909.html
Fixed: 3312f2d21f13 ("Cygwin: console: Redesign mode set strategy on close().")
Reported-by: Jeremy Drake <cygwin@jdrake.com>
Cherry-picked-from: dc6aa87


dscho (approved):
What a delightfully clear commit message and obvious diff!

PR #280, opened 2025-04-18
CI: run msys2-tests against freshly built runtime artifacts

### LINKED ISSUE #24: Figure out what to do with msys2-runtime tests in this repo
It has a copy of the [convert code](https://github.com/msys2/msys2-runtime/blob/msys2-3_3_6-release/winsup/cygwin/msys2_path_conv.cc) and a bunch of [tests](https://github.com/msys2/path_convert/blob/master/src/main.cpp#L169).

Suggestion:

* [ ] Make sure the code is in sync with the runtime
* [x] Test against the real runtime somehow, if all fails via a dummy mingw binary that just prints all args and env vars back https://github.com/msys2/msys2-tests/pull/21
* [x] Move the testing code to https://github.com/msys2/msys2-tests, so we run it more often, against staging, and can potentially add more runtime tests https://github.com/msys2/msys2-tests/pull/22
* ~Archive this repo~

jeremyd2019:
Call msys2-tests composite workflows to run tests using built artifacts.

Fixes msys2/msys2-tests#24

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,6 +2,9 @@ name: build
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -34,3 +37,57 @@ jobs:
         with:
           name: install
           path: _dest/
+
+  generate-test-matrix:
```
dscho: For clarity, this could be named `generate-msys2-tests-matrix`.

.github/workflows/build.yaml
```diff
@@ -34,3 +37,57 @@ jobs:
...
+
+      - name: Add staging repo
+        shell: msys2 {0}
+        run: |
+          sed -i '1s|^|[staging]\nServer = https://repo.msys2.org/staging/\nSigLevel = Never\n|' /etc/pacman.conf
+
+      - name: Update using staging
+        run: |
+          msys2 -c 'pacman --noconfirm -Suuy'
+          msys2 -c 'pacman --noconfirm -Suu'
```
dscho: Ah, I just wondered why these do not tot to be in two _separate_ steps in case, say, the `msys2-runtime` is updated in the first `pacman` call: The default shell on Windows is PowerShell, which is used here.

Maybe we want an explicit `shell: pwsh` here?

Also, unless I am mistaken, an error in `pacman` invocation would not be caught unless these two lines (or something equivalent) was added:

```powershell
$ErrorActionPreference = 'Stop'
$PSNativeCommandUseErrorActionPreference = $true
```

dscho (approved):
I like the direction!

Below, I only offer minor suggestions that I believe would improve the PR, but I won't insist on them being implemented unless you agree that they are worth it.

.github/workflows/build.yaml
jeremyd2019: I copied that from msys2-tests.  I expect that it *wants* to ignore the error, because when pacman terminates things in a core update it shows up as an error.  I have used:
```yaml
      - name: Update from staging
        continue-on-error: true
        run: |
          pacman --noconfirm --noprogressbar --overwrite '*' -Syuu
      - name: Update from staging 2
        run: |
          pacman --noconfirm --noprogressbar --overwrite '*' -Syuu
```

dscho: I guess that could work. I am always a bit fuzzy on the exact logic of `continue-on-error`: I _think_ it mistakes _any_ output to `stderr` to mean that there was an error. But that would probably also apply to my suggested change, so 🤷

dscho (approved):

PR #286, opened 2025-06-03
Do retrieve AzureAD users' information again

dscho:
As of the upgrade to v3.6.2, I can no longer open any MinTTY windows:

image
This symptom (the message "Error: Could not fork child process: There are no available terminals (-1)") was also [described on the Cygwin mailing list](https://inbox.sourceware.org/cygwin/CAL148o6i=r+G=cEQzs5e+KmMt07FxJHhMnoNnquhM0JJuyrwtA@mail.gmail.com/#t).

I've bisected the problem down to 48e7d63268 (Cygwin: fetch_account_from_windows: skip LookupAccountSid for SIDs known to fail, 2025-04-10), which prevents my AzureAD user account (the SID begins with `S-1-12-1-`) from being handled correctly.

This PR is a (minimally) partial revert of that commit to resolve this problem.


lazka:
https://github.com/msys2/MSYS2-packages/pull/5434

PR #292, opened 2025-06-20
Fix creating native symlinks pointing to `..`

dscho:
Git's test suite currently does not pass under `MSYS=winsymlinks:nativestrict`. Most of those issues need to be addressed in Git, as they are either due to incorrect assumptions, or due to bugs in the symlink support code.

However, one of the issues is in a test that first creates a symlink pointing to `..` and later verifies that following that symlink in `cat-file` works: Instead of creating a link to `..`, `ln -s` creates a link to `../../$(basename "$PWD")`. This link still points to the same entity, but it is incorrect, and not due to a bug in Git, but due to a bug in the MSYS2 runtime.

This fixes t1006.311(git cat-file --batch-check --follow-symlinks works for .. links).

I contributed this fix to the Cygwin project here: https://inbox.sourceware.org/cygwin-patches/6058889e2ae8c9c827a8d6678f09b3b1741e2fcf.1750413578.git.johannes.schindelin@gmx.de/T/#u. This PR is a companion of https://github.com/git-for-windows/msys2-runtime/pull/101.

### DIFF
--- a/winsup/cygwin/path.cc
+++ b/winsup/cygwin/path.cc
@@ -2030,9 +2030,18 @@ symlink_native (const char *oldpath, path_conv &win32_newpath)
       while (towupper (*++c_old) == towupper (*++c_new))
 	;
       /* The last component could share a common prefix, so make sure we end
-         up on the first char after the last common backslash. */
-      while (c_old[-1] != L'\\')
-	--c_old, --c_new;
+         up on the first char after the last common backslash.
+
+	 However, if c_old is a strict prefix of c_new (at a component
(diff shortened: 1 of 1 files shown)


jeremyd2019:
I noticed this behavior, but I somehow thought it was due to trying to deal with some limitation of Windows symlinks or trying to replicate some obscure corner-case of unix symlink behavior, rather than a bug in finding the common prefix

dscho:
➡️ https://github.com/msys2/MSYS2-packages/pull/5468

PR #298, opened 2025-07-05
Fix yet another Ctrl+C bug

dscho:
This fixes the bug where <kbd>Ctrl</kbd>+<kbd>C</kbd> after launching `cat | /c/windows/system32/ping -t localhost` in a PowerShell session in a Windows Terminal window would terminate the `ping` process but not the `cat` process.

This commit is cherry-picked from the `cygwin-3_6-branch`.

To increase the confidence in this bug fix, I [added some integration tests to Git for Windows' fork of the MSYS2 runtime](https://github.com/git-for-windows/msys2-runtime/pull/105), including the reproducer for the bug corresponding to this here fix.


lazka:
I guess this can be closed after #304 ?

dscho:
> I guess this can be closed after #304 ?

Yes!

Issue #297, opened 2025-07-04
Please consider POSIXifying the `GNUPGHOME` variable


chawyehsu:
### Context

`GNUPGHOME` is an environment variable used by gnupg, which allows users to change the home directory of gpg. _gnupg for Windows_ is an official build that supports changing the home directory using `GNUPGHOME` in the form of `C:\Users\user\.gnupg`. In the community however, some tools such as git-for-windows, package a gnupg build based on MSYS2 and rely on the msys2-runtime and only supports parsing and using `GNUPGHOME` in the form of `/c/Users/user/.gnupg` to change the home directory.

### Why

Under certain conditions, users may actively or passively install multiple editions of gnupg, including the two different gnupg mentioned above. When they want to share the same gnupg home directory across different gnupg builds, they may use the `GNUPGHOME` variable. However, due to different interpretations of the `GNUPGHOME` variable, this goal cannot be achieved. This has been discussed in the gnupg mailing list [[1]][[2]], and similar issues have also been reported in the git-for-windows issue tracker [[3]][[4]]. Specifically, when users set `GNUPGHOME=C:\Users\user\.gnupg` or `GNUPGHOME=C:/Users/user/.gnupg`, the gnupg based on msys2-runtime will be unable to use this variable and will report an error.

```
$ gpgconf.exe -L
sysconfdir:/etc/gnupg
bindir:/usr/bin
libexecdir:/usr/lib/gnupg
libdir:/usr/lib/gnupg
datadir:/usr/share/gnupg
localedir:/usr/share/locale
socketdir:C%3a\Users\user\.gnupg
dirmngr-socket:C%3a\Users\user\.gnupg/S.dirmngr
keyboxd-socket:C%3a\Users\user\.gnupg/S.keyboxd
agent-ssh-socket:C%3a\Users\user\.gnupg/S.gpg-agent.ssh
agent-extra-socket:C%3a\Users\user\.gnupg/S.gpg-agent.extra
agent-browser-socket:C%3a\Users\user\.gnupg/S.gpg-agent.browser
agent-socket:C%3a\Users\user\.gnupg/S.gpg-agent
homedir:C%3a\Users\user\.gnupg
```

### Proposal

Propose to POSIXify the `GNUPGHOME` variable, similar to the way of handling `SHELL` implemented by @dscho in [[5]]. Such a proposal was also mentioned in the comments [[6]]. I can draft a pull-request or patch if this is reasonable.

[...]

PR #300, opened 2025-07-09
POSIX-ify the GNUPGHOME variable

chawyehsu:
When GnuPG is built for MSYS2 (e.g., as part of Git for Windows), it expects
the `GNUPGHOME` environment variable to contain a POSIX-style path
(e.g., `/c/Users/user/.gnupg`). However, users often set this variable using a
Windows-style path (e.g., `C:\Users\user\.gnupg`), which is the native format
for GnuPG for Windows.

This discrepancy leads to MSYS2-based GnuPG failing to correctly locate its home
directory when running within an environment having the `GNUPGHOME` defined in
the Windows-style, as the Windows path is not properly interpreted. This issue
prevents users from seamlessly sharing a single GnuPG home directory across
different GnuPG installations (native Windows build vs. MSYS2-based). [1] [2]

To resolve this, we posix-ify `GNUPGHOME` in a manner similar to `SHELL`. [3]
This change will allow MSYS2-based GnuPG to correctly interpret the `GNUPGHOME`
variable regardless of whether it's set with a Windows path, improving the
interoperability and user experience.

Issue [4] and [5] from git-for-windows should benefit from this change.

This addresses https://github.com/msys2/msys2-runtime/issues/297

[1]: https://lists.gnupg.org/pipermail/gnupg-users/2025-February/067533.html
[2]: https://lists.gnupg.org/pipermail/gnupg-users/2025-March/067538.html
[3]: https://github.com/msys2/msys2-runtime/commit/8bb4359ac2b3866467697e69e0142341b00e70b1
[4]: https://github.com/git-for-windows/git/discussions/5607
[5]: https://github.com/git-for-windows/git/issues/3857

### DIFF
--- a/winsup/cygwin/environ.cc
+++ b/winsup/cygwin/environ.cc
@@ -323,6 +323,7 @@ static win_env conv_envvars[] =
   {
     {NL ("PATH="), NULL, NULL, env_PATH_to_posix, env_plist_to_win32, true},
     {NL ("HOME="), NULL, NULL, env_path_to_posix, env_path_to_win32, false},
+    {NL ("GNUPGHOME="), NULL, NULL, env_path_to_posix, env_path_to_win32, false},
     {NL ("LD_LIBRARY_PATH="), NULL, NULL,
 			       env_plist_to_posix, env_plist_to_win32, true},
     {NL ("ORIGINAL_PATH="), NULL, NULL, env_PATH_to_posix, env_plist_to_win32, true},
@@ -349,7 +350,7 @@ static const unsigned char conv_start_chars[256] =
     0,        0,        0,        0,        0,        0,        0,        0,
     0,        0,        0,        0,        0,        0,        0,        0,
 /*            A         B         C         D         E         F         G */
-    0,        0,        0,        0,        0,        0,        0,        0,
+    0,        0,        0,        0,        0,        0,        0,        WC,
     /*  72 */
 /*  H         I         J         K         L         M         N         O */
     WC,       0,        0,        0,        WC,       0,        0,        WC,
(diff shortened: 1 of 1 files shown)


jeremyd2019:
Has anyone considered that win32 gpg and cygwin gpg won't be able to talk to each other?  cygwin will be using their unix sockets, and has its own pid namespace, so it seems likely that you could get two different gpg-agents and/or dirmngrs running on the same GNUPGHOME, and I don't know what sort of issues that could cause.

dscho:
> win32 gpg and cygwin gpg won't be able to talk to each other

That's a really good point I had not considered before.

My thinking involved the keyring only, which _should_ be able to be shared. @chawyehsu could you test what happens when you start a Win32 GPG agent and try to use it via the MSYS2 GPG, and also what happens when you start an MSYS2 GPG agent and try to use it via Win32 GPG? (You should be able to perform these tests by installing `mingw-w64-x86_64-gnupg` and `gnupg` side by side.)

jeremyd2019:
I seem to remember that mingw-w64-gnupg didn't actually work right... probably better to use gpg4win.org (or the win32 package of gnupg)

I was thinking, maybe gnupg needs the same sort of treatment as ssh (do you want to use the msys2 version or the win32 version?), though I hate the idea of yet another GfW installer option

mlt:
My major grievance with dual gpg is that gpg-agent blocks access to a smart card so another one can't access it and they indeed do not talk to each other. I strongly prefer using native gpg4win in general. That way I can add it to the PATH so Thunderbird can find gpgme instead of adding everything in msys2 to PATH, also Kleopatra is there. But! I do need `/usr/bin/gpg` if I want to sign custom repo.
Maybe it would be better if there was a way for a user to choose binary to use via env var and tools like `repo-add` considered it.

PR #301, opened 2025-07-14
[backport] stdin flush fix

### LINKED ISSUE #199: Flushing stdin doesn't seem to work
This is an example extracted from pacman https://gitlab.archlinux.org/pacman/pacman/-/blob/2180e4d127af2ce0722b831a7131283604a7bdf1/src/pacman/util.c#L185

pacman flushes stdin before asking a question and doesn't have and takes any non-"y" as not, leading to user confusion if for example enter is pressed during the waiting before the question is asked, leading to surprise aborts of the process, example: https://github.com/msys2/MSYS2-packages/issues/3532

This example allows you to input text, then flushes as pacman does, then reads stdin. Flushing currently doesn't have an effect.

```c
#include <stdio.h>
#include <unistd.h>
#include <termios.h>

int main() {
    printf("type something in the next 5 secs\n");
    for (int i=0; i < 5; i++) {
        sleep(1);
        printf("%d\n", i);
    }

    // Flush stdin
    int fd = fileno(stdin);
    if (fd != -1) {
        printf("flushing stdin\n");
        if (tcflush(fd, TCIFLUSH) == -1) {  // this doesn't seem to do anything
            perror("Error flushing stdin");
            return 1;
        }
    } else {
        perror("Error getting file descriptor for stdin");
        return 1;
    }

    char buffer[100];
    printf("Enter some text:\n");
    fgets(buffer, sizeof(buffer), stdin);
    printf("You entered: %s\n", buffer);

    return 0;
}
```

lazka:
from 2aa41b516055ea9383508342706288deb3baf1f6

Fixes #199

### DIFF
--- a/winsup/cygwin/fhandler/pty.cc
+++ b/winsup/cygwin/fhandler/pty.cc
@@ -1613,9 +1613,15 @@ fhandler_pty_slave::tcflush (int queue)
   if (queue == TCIFLUSH || queue == TCIOFLUSH)
     {
       char pipe[MAX_PATH];
+#ifdef __MSYS__
+      __small_sprintf (pipe,
+		       "\\\\.\\pipe\\msys-%S-pty%d-master-ctl",
+		       &cygheap->installation_key, get_minor ());
+#else
       __small_sprintf (pipe,
 		       "\\\\.\\pipe\\cygwin-%S-pty%d-master-ctl",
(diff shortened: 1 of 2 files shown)


dscho (approved):

dscho:
↦ https://github.com/msys2/MSYS2-packages/pull/5513

PR #303, opened 2025-07-14
fixup! Add MSYS2 triplet

### LINKED ISSUE #302: Drop "Add MSYS2 triplet" patch
Currently f51657ba5f587a296c.

We now use "x86_64-pc-cygwin" in MSYS2, so this commit is no longer doing anything, and removing it would simplify rebasing if upstream rewrites those files. There is the one hunk which changes the dll prefix from "cyg" to "msys-", but the runtime doesn't have any DLLs which this affects currently. We could move that change to the cygwin branch, or do that when/if needed.

dscho:
We now use "x86_64-pc-cygwin" in MSYS2, so this commit is no longer doing anything, and removing it would simplify rebasing if upstream rewrites those files. There is the one hunk which changes the dll prefix from "cyg" to "msys-", but the runtime doesn't have any DLLs which this affects currently.

This fixes https://github.com/msys2/msys2-runtime/issues/302.


lazka:
oh, thanks. lgtm

dscho:
↳ https://github.com/msys2/MSYS2-packages/pull/5514

PR #312, opened 2025-09-21
[backport] Cygwin: lock cygheap during fork

lazka:
As suggested here: https://github.com/msys2/MSYS2-packages/pull/5637#issuecomment-3315326598

### DIFF
--- a/winsup/cygwin/fork.cc
+++ b/winsup/cygwin/fork.cc
@@ -329,6 +329,7 @@ frok::parent (volatile char * volatile stack_here)
   /* NEVER, EVER, call a function which in turn calls malloc&friends while this
      malloc lock is active! */
   __malloc_lock ();
+  cygheap->lock ();
   bool locked = true;
 
   /* Remove impersonation */
@@ -483,6 +484,7 @@ frok::parent (volatile char * volatile stack_here)
 		   impure, impure_beg, impure_end,
 		   NULL);
 
+  cygheap->unlock ();
   __malloc_unlock ();
   locked = false;
   if (!rc)
@@ -568,7 +570,10 @@ cleanup:
   if (fix_impersonation)
     cygheap->user.reimpersonate ();
   if (locked)
-    __malloc_unlock ();
+    {
+      cygheap->unlock ();
+      __malloc_unlock ();
+    }
 
   /* Remember to de-allocate the fd table. */
   if (hchild)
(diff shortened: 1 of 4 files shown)


jeremyd2019 (approved):
LGTM, and will make rust happier at least. (this won't help the "resource unavailable" fork errors, but will help hangs when building with lots of threads)

lazka:
So for the other errors retrying might work (I tried 3 times so far for the msys build)?

jeremyd2019:
No I don't think retrying will help, the base addresses are deterministic.  I think you'd need to try different paths (I think that's why it succeeded in MSYS2 CI (`/d/_/rust`) and not autobuild (`/d/S/B`)).  If you want to try to get it to work in autobuild, you could try changing the paths inside the `$srcdir`, like renaming the rustc-1.90.0-src directory, or adding `build-dir = "B"` or other values to the bootstrap.toml (in the `[build]` section)

lazka:
Ah, hm, that. I was wondering if we should patch binutils to take a seed from an env var instead (is that binutils??). The python update is also stuck on address conflicts..

jeremyd2019:
Yeah, binutils - circa https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=ld/emultempl/pep.em;h=60a833947bd524d06241d9d54f5999f69594643d;hb=HEAD#l937 for 64-bit (https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=ld/emultempl/pe.em;h=b522687c1d5b4e4fb1d3d998b6ee5f19232b264c;hb=HEAD#l968 for 32-bit).

The posix_spawn work I was doing helps with rust, because it uses posix_spawn to call the linker.  Those changes result in posix_spawn not using a fork, so it doesn't matter that the dll can't load to the same address.  I don't know if python uses posix_spawn or fork/exec directly so I don't know if that would help it.

lazka:
It should in some cases: https://github.com/python/cpython/pull/113118 though cygwin might need 
posix_spawn_file_actions_addclosefrom_np for this

jeremyd2019:
no closefrom in cygwin, so not in posix_spawn either

dscho:
Hrm. It seems that this change causes pretty consistent CI failures over in Cygwin: https://github.com/cygwin/cygwin/actions/workflows/cygwin.yml

jeremyd2019:
no, it was the switch of windows-latest from 2022 to 2025: https://github.com/cygwin/cygwin/commit/d704d04a2ab61e92e105ba99d25df3f52ee653ff

PR #329, opened 2026-03-12
Cygwin: Fix segfault when XSAVE area sizes are unaligned

lazka:
During signal delivery, Cygwin saves the CPU's extended register state
(floating-point, SSE, AVX, etc.) to a stack buffer using the xsave64
instruction, which requires its destination to be 64-byte aligned.
Before executing xsave64, the code queries the CPU (via cpuid) for the
required buffer size, then subtracts that size (plus a fixed overhead)
from the stack pointer.

The stack alignment arithmetic assumes that cpuid returns a size that
is a multiple of 64. Until recently, this held true for all x86 CPUs.
On recent AMD and Intel CPUs, however, the PKU feature (Protection Keys
for Userspace, a memory-protection mechanism) adds an XSAVE component
of only 8 bytes, which makes the total size no longer a multiple of 64.
The subtraction then places the xsave64 buffer at a misaligned address,
causing a segfault.

This was first observed when running Cygwin/MSYS2 under Wine on Linux,
where the host kernel exposes the PKU feature directly. The same
problem could surface on future Windows versions that expose PKU or
other small XSAVE components.

The fix rounds up the cpuid-reported size to the next 64-byte multiple
before using it in the stack allocation. The existing code already
guarantees correct alignment for any buffer size that is a multiple of
64, so this rounding is sufficient.

Fixes: c607889824 ("Cygwin: sigfe: Fix a bug that signal handler destroys fpu states")
Signed-off-by: Pip Cet <pipcet@protonmail.com>

Source: https://cygwin.com/pipermail/cygwin/2025-June/258375.html

### DIFF
--- a/winsup/cygwin/scripts/gendef
+++ b/winsup/cygwin/scripts/gendef
@@ -233,6 +233,8 @@ sigdelayed:
 	xorl	%ecx,%ecx
 	cpuid	# get necessary space for xsave
 	movq	%rbx,%rcx
+	addq	\$63, %rbx
+	andq	\$-64, %rbx # align to next 64-byte multiple
 	addq	\$0x48,%rbx # 0x18 for alignment, 0x30 for additional space
 	subq	%rbx,%rsp
 	movl	%ebx,0x24(%rsp)


jhol:
Related Wine bug: https://bugs.winehq.org/show_bug.cgi?id=58420 .

dscho (commented):
I found that the commit message does not really manage to "pick me up where I'm at", and I had to do quite a bit of research (with a bit of help of Claude Opus 4.6) to gather the context necessary to understand it. For that reason, I would like to propose this commit message instead:

```
   Cygwin: Fix segfault when XSAVE area sizes are unaligned

   During signal delivery, Cygwin saves the CPU's extended register state
   (floating-point, SSE, AVX, etc.) to a stack buffer using the xsave64
   instruction, which requires its destination to be 64-byte aligned.
   Before executing xsave64, the code queries the CPU (via cpuid) for the
   required buffer size, then subtracts that size (plus a fixed overhead)
   from the stack pointer.

   The stack alignment arithmetic assumes that cpuid returns a size that
   is a multiple of 64. Until recently, this held true for all x86 CPUs.
   On recent AMD and Intel CPUs, however, the PKU feature (Protection Keys
   for Userspace, a memory-protection mechanism) adds an XSAVE component
   of only 8 bytes, which makes the total size no longer a multiple of 64.
   The subtraction then places the xsave64 buffer at a misaligned address,
   causing a segfault.

   This was first observed when running Cygwin/MSYS2 under Wine on Linux,
   where the host kernel exposes the PKU feature directly. The same
   problem could surface on future Windows versions that expose PKU or
   other small XSAVE components.

   The fix rounds up the cpuid-reported size to the next 64-byte multiple
   before using it in the stack allocation. The existing code already
   guarantees correct alignment for any buffer size that is a multiple of
   64, so this rounding is sufficient.

   Fixes: c607889824 ("Cygwin: sigfe: Fix a bug that signal handler destroys fpu states")
   Signed-off-by: Pip Cet <pipcet@protonmail.com>
```

The patch itself looks correct to me, and I could not find any more minimal solution (there isn't any x86_64 assembly instruction to add _and_ round up in one go.

lazka:
thank. I've updated the commit message with your proposal.

dscho:
I'll open the corresponding MSYS2-packages PR in a bit.

lazka:
I missed to update the default branch here, so this was merged into 3.6.6 instead of 3.6.7. I've cherry-picked the commit to 3.6.7 now and updated the default branch.

leuasseurfarrelds247-arch (approved):

Issue #330, opened 2026-03-25
rsync -e flag causes path misread with /c/ paths (source and destination cannot both be remote)


vicjayjay:
## Description

When using MSYS2 rsync (rsync 3.4.1 from Git for Windows) with the `-e` flag and a remote destination containing an MSYS2-style path (e.g., `/c/Users/...`), rsync fails with:

```
rsync error: syntax or usage error (code 1) at main.c(1415) [Receiver=3.4.1] - The source and destination cannot both be remote.
```

This happens because MSYS2 path conversion converts the `/c/` portion of the remote destination path (`user@host:/c/path`) and rsync then misinterprets the remote path as a local path, triggering the error.

## Steps to Reproduce

On a Windows machine with Git for Windows (MSYS2 rsync):

```bash
# FAILS with 'source and destination cannot both be remote':
rsync -az -e "ssh -p 2222" "/c/Users/test/file.txt" "user@host:/c/Users/test/"

# WORKS correctly (using RSYNC_RSH env var instead of -e flag):
RSYNC_RSH="ssh -p 2222" rsync -az "/c/Users/test/file.txt" "user@host:/c/Users/test/"
```

Note: The remote side must have rsync installed (e.g., cwrsync) for the connection to succeed once the path bug is bypassed.

## Expected Behavior

Both invocation styles (`-e flag` and `RSYNC_RSH` env var) should behave identically. The `-e` flag should pass the remote path unchanged to SSH/rsync on the remote side.

## Root Cause Analysis

The issue appears to be in how MSYS2's path conversion layer handles arguments when rsync spawns the remote shell via `-e`. When the remote path argument containing `/c/` (MSYS2 Windows drive path format) is passed through the MSYS2 runtime's `spawn()` or `exec()` call, the path conversion converts the `/c/` portion before rsync can pass it to the SSH process unchanged. This causes rsync to misinterpret `user@host:/c/path` as having a local path component.

## Workaround

Use `RSYNC_RSH` environment variable instead of the `-e` flag:

```bash
export RSYNC_RSH="ssh -p 2222 -i /c/Users/.ssh/id_rsa"
rsync -az /c/Users/local/path/ user@host:/c/Users/remote/path/
```

## Environment

- OS: Windows 10/Server (MSYS2/Git for Windows)
- rsync: 3.4.1 (from Git for Windows bundled MSYS2)
- SSH: OpenSSH_for_Windows_9.x
- MSYS2 runtime version: (please fill in)

[...]

PR #331, opened 2026-03-25
fix: skip MSYS2 path conversion for rsync-over-SSH remote paths

vicjayjay:
## Summary

When using MSYS2 rsync with the `-e ssh` flag and a remote destination containing an MSYS2-style path (e.g., `user@host:/c/path`), rsync fails with:

```
rsync error: syntax or usage error (code 1) at main.c(1415) - The source and destination cannot both be remote.
```

## Root Cause

In `arg_heuristic_with_exclusions()` (winsup/cygwin/path.cc), the MSYS2 path conversion heuristic treats the `/c/` portion of `user@host:/c/path` as a Windows drive-letter path and converts it. The converted path is then passed to SSH, causing rsync on the remote side to misparse the path.

## The Fix

Added a check for remote-path syntax before invoking the conversion:

```c
const char *at_sign = strchr(arg, '@');
const char *colon   = strchr(arg, ':');
if (at_sign && colon && at_sign < colon)
  return (char*)arg;
```

This reliably identifies SCP/rsync remote-path syntax (`user@host:/path`) and skips path conversion for those arguments. The `@:` pattern is unique to remote-path syntax — a Windows drive letter never has `@` before `:`.

## Why This Fix Is Correct

| Argument pattern | @ before : ? | Action |
|---|---|---|
| `/c/Users/file.txt` | NO | Convert (local MSYS2 path) |
| `user@host:/c/path` | YES | SKIP conversion (remote path) |
| `ssh://host/path` | NO | Not affected (already skipped) |

## Testing

**Before fix:**
```bash
rsync -az -e "ssh -p 2222" "/c/Users/test/" "user@host:/c/backup/"
# rsync error: source and destination cannot both be remote
```

**After fix:**
```bash
rsync -az -e "ssh -p 2222" "/c/Users/test/" "user@host:/c/backup/"
# Works correctly
```

**Regression check:** Local MSYS2 paths (`/c/Users/...`) continue to be converted normally.

## Issue

Follow-up to: #330

### DIFF
--- a/winsup/cygwin/msys2_path_conv.cc
+++ b/winsup/cygwin/msys2_path_conv.cc
@@ -364,6 +364,12 @@ skip_p2w:
     if (*it == ':')
         goto skip_p2w;
 
+    // Track '@' position for SCP/rsync remote path detection.
+    // Only set when the prefix looks like a valid username
+    // (alphanumeric, dots, dashes).
+#define ALNUMDD ".-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+    const char *at = NULL;
+
     while (it != end && *it) {
         switch (*it) {
         case '`':
@@ -389,11 +395,24 @@ skip_p2w:
                 if (it + 3 < end && it[2] == '.' && it[3] == '/')
                     goto skip_p2w;
             }
+
+            // Skip SCP/rsync remote paths: user@host:/path
+            // Verify: at was set, hostname between at+1 and colon
+            // is all valid hostname chars, and colon is followed
+            // by / then an alphanumeric character.
+            if (at && strspn(at + 1, ALNUMDD) == (size_t)(it - at - 1)
+                && it + 2 < end && it[1] == '/' && isalnum(it[2]))
+                goto skip_p2w;
             break;
         case '@':
             // Paths do not contain '@@'
             if (it + 1 < end && it[1] == '@')
                 goto skip_p2w;
+
+            // Record '@' only if the prefix is a valid username
+            if (!at && strspn(*src, ALNUMDD) == (size_t)(it - *src))
+                at = it;
+            break;
         }
         ++it;
     }


dscho:
(I fixed the PR base branch, it was targeting `master` instead of the current default branch, `msys2-3.6.7`.)

winsup/cygwin/path.cc
```diff
@@ -4187,6 +4187,19 @@ arg_heuristic_with_exclusions (char const * const arg, char const * exclusions,
...
+     drive-letter path and gets incorrectly converted, causing rsync to
+     report "source and destination cannot both be remote".
+     The presence of '@' before ':' is a reliable indicator that this is
+     a remote path (SCP/rsync syntax: user@host:/path) and not a Windows
+     drive letter. */
+  const char *at_sign = strchr(arg, '@');
+  const char *colon   = strchr(arg, ':');
+  if (at_sign && colon && at_sign < colon)
+    return (char*)arg;
+
```
dscho: You _almost_ found the correct spot where this new logic needs to live, but it's not quite here. 11 lines further down, the [`convert()`](https://github.com/vicjayjay/msys2-runtime/blob/dd57a10cfecdee0773cdee187b0fa6cd31a94a56/winsup/cygwin/msys2_path_conv.cc#L263-L326) function is called, _that_ is closer to the correct location. The actual location where things like this need to happen is [the `find_path_start_and_type()` function](https://github.com/vicjayjay/msys2-runtime/blob/dd57a10cfecdee0773cdee187b0fa6cd31a94a56/winsup/cygwin/msys2_path_conv.cc#L339-L539), I think.

If you look carefully, you will see that this function avoids performance hitters like `strchr()`.

You might find inspiration in https://github.com/msys2/msys2-runtime/pull/182/changes/afa529d7dcfa0713417f6c71785da1ce4a994fee how this could be done in a better way. For example, you could initialize `char *colon = NULL;` just before [the loop](https://github.com/vicjayjay/msys2-runtime/blob/dd57a10cfecdee0773cdee187b0fa6cd31a94a56/winsup/cygwin/msys2_path_conv.cc#L367), then set `colon = it;` in [the `case ':'` arm](https://github.com/vicjayjay/msys2-runtime/blob/dd57a10cfecdee0773cdee187b0fa6cd31a94a56/winsup/cygwin/msys2_path_conv.cc#L380), and then add an `if (!colon) goto skip_p2w;` in [the `case '@': arm](https://github.com/vicjayjay/msys2-runtime/blob/dd57a10cfecdee0773cdee187b0fa6cd31a94a56/winsup/cygwin/msys2_path_conv.cc#L393).

But even so, that logic might very well catch _too much_, i.e. by mistake prevent legitimate paths from being converted. `@` characters are [allowed in filenames](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#:~:text=Use%20any%20character%20in%20the%20current%20code%20page%20for%20a%20name), after all, and colons can be path separators.

And even if the "`@` before `:`" pattern is probably to broad, at the same time it is _also_ too narrow. You could, after all, have a section like this in your `~/.ssh/config` (I have plenty of those in mine):

```
Host my-machine
HostName xyz.example.com
User vicjayjay
```

And then you can call `rsync abc my-machine:/c/xyz/`. In fact, I rarely ever specify my user name explicitly in any SSH address, I pretty much always use those nicknames defined in `~/.ssh/config`. And you'll note that your "`@` before `:`" pattern does _not_ handle this situation well...

vicjayjay:
@dscho Thanks for the detailed review — updated in bdc781d.

**Changes:**
- Moved the logic from `arg_heuristic_with_exclusions()` in `path.cc` into `find_path_start_and_type()` in `msys2_path_conv.cc`
- Replaced `strchr()` scans with in-loop tracking: `colon` is set on the first `':'` encountered, then in `case '@'` we check `if (!colon) goto skip_p2w;`
- Removed the old block from `path.cc` entirely

**Limitations acknowledged in the commit message:**
- SSH config aliases (`my-machine:/path` without `@`) are not covered — we can't reasonably parse `~/.ssh/config` from here
- `@` in filenames could theoretically false-positive, though `@` before the first `:` in a real Windows/MSYS2 path seems unlikely in practice

Happy to refine further if you have a better approach for the alias case.

vicjayjay:
CI note: the 4 failed jobs are unrelated to this change:

- **CLANG64-clang, UCRT64-clang, CLANGARM64-clang**: `error: target not found: mingw-w64-clang-aarch64-openmp` — the `openmp` package appears to have been removed from the MSYS2 repos recently. All gcc test variants pass.
- **MSYS-gcc**: GitHub Actions network timeout (`api.github.com:443` unreachable during `setup-msys2` download) — transient infrastructure issue.

The build itself succeeded (10m40s) and all gcc-based test jobs (UCRT64-gcc, MINGW64-gcc, MINGW32-gcc) pass cleanly.

dscho:
> * **CLANG64-clang, UCRT64-clang, CLANGARM64-clang**: `error: target not found: mingw-w64-clang-aarch64-openmp` — the `openmp` package appears to have been removed from the MSYS2 repos recently. All gcc test variants pass.

That's indeed a regression. More precisely, the `openmp` package never existed as such, only as an alias for `llvm-openmp` and that one was (accidentally?) removed. I opened https://github.com/msys2/msys2-tests/pull/94 to fix that. Once that's merged, we can re-run.

> * **MSYS-gcc**: GitHub Actions network timeout (`api.github.com:443` unreachable during `setup-msys2` download) — transient infrastructure issue.

Yes, this was transient. GitHub seems to have a rough time lately. I re-ran it, and it passed.

winsup/cygwin/msys2_path_conv.cc
```diff
@@ -389,11 +391,21 @@ path_type find_path_start_and_type(const char** src, int recurse, const char* en
...
+                colon = it;
             break;
         case '@':
             // Paths do not contain '@@'
             if (it + 1 < end && it[1] == '@')
                 goto skip_p2w;
+
+            // Skip SCP/rsync remote paths (user@host:/path).
+            // If '@' appears before any ':', the argument likely
+            // uses remote-shell syntax and must not be converted.
```
dscho: I am still worried that this might mess with path _lists_, i.e. colon-separated values in the form as e.g. `PATH`.

Besides, we're not even ensuring that there _is_ a colon here, so even `abc@` would match, or `@abc`, which is clearly unwanted.

Personally, I'd try to match a pattern like this: [sequence-of-alnum-dashes-and-dots]@[sequence-of-alnum-dashes-and-dots]:/[alnum]. That might be overkill, but it would be a lot less likely to introduce inadvertent side effects.

And to be clear: I would not implement this as a state machine, even if that would be the fastest, performance-wise. I would probably introduce a `char *at = NULL` that, before it is assigned, runs something like:

```c
// Does the string so far consist only of alphanumerical characters, dots or dashes,
// i.e. does it look like a username?
#define ALNUMDD ".-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
if (strspn(src, ALNUMDD) == it)
  at = it;
```

and then, in the `:` arm, if this is the first `:` and `at != NULL` (i.e. the string started with something like `<username>@`) do the same dance with `strspn(at + 1, ALNUMDD)` to verify the host part, and then verify the path part via: `it[1] == '/' && isalnum(it[2])`.

Oh, and I think this needs an addition to: https://github.com/msys2/msys2-tests/blob/main/runtime/path_convert/src/main.cpp

vicjayjay:
@dscho Updated in 1ffd11f with the stricter pattern you suggested.

**What changed:**

The loose `!colon` check is replaced with strict `strspn()`-based validation:

```c
#define ALNUMDD ".-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
const char *at = NULL;

// In case '@': record only if prefix is valid username chars
if (!at && strspn(*src, ALNUMDD) == (size_t)(it - *src))
    at = it;

// In case ':': skip only if hostname is valid AND colon is followed by /[alnum]
if (at && strspn(at + 1, ALNUMDD) == (size_t)(it - at - 1)
    && it + 2 < end && it[1] == '/' && isalnum(it[2]))
    goto skip_p2w;
```

This matches the pattern `[alnum-dot-dash]@[alnum-dot-dash]:/[alnum]` — so:
- `user@host:/c/path` → skipped (valid SCP syntax)
- `abc@` → not skipped (`at` never set, no colon)
- `@abc` → not skipped (empty prefix fails `strspn` check)
- `file@name` with no colon → not skipped
- `C:/path@name:/foo` → not skipped (prefix contains `/`, fails username check)
- PATH-style lists like `/usr/bin:/usr/local/bin` → not skipped (no `@`)

Re: tests — I'll add cases to `msys2-tests/runtime/path_convert/src/main.cpp` in a follow-up or as part of this PR if you prefer. Let me know.

Thanks for the `openmp` fix (msys2/msys2-tests#94) — once that merges the CI should go green.

dscho:
@vicjayjay wow that was fast. Did you use an AI agent or what 😆

dscho:
> Re: tests — I'll add cases to `msys2-tests/runtime/path_convert/src/main.cpp` in a follow-up or as part of this PR if you prefer. Let me know.

Since those tests live in a different repository, I prefer the changes in a parallel PR ;-)

mati865:
> @vicjayjay wow that was fast. Did you use an AI agent or what 😆 

Yep, you are basically talking with an openclaw agent the whole time.

Kreijstal:
problem with these heuristics is that shit goes down fast, not sure about this, can paths have.

Also just tell the agent to use MSYS2_ENV_CONV_EXCL and MSYS2_ARG_CONV_EXCL the agent will stop caring, use the workaround and never come back. Agents love workarounds

PR #333, opened 2026-03-27
Cygwin: make create_root_entry tolerant of concurrent mount table initialization

ilude:
## Problem

When multiple MSYS2 processes start concurrently, `create_root_entry()`
can crash with:

    fatal error - add_item ("\??\C:\Program Files\Git", "/", ...) failed, errno 1

The `add_item()` call returns `EPERM` because it finds an existing
immutable `/` mount entry and `MOUNT_OVERRIDE` is not set.

This has been reported in several contexts where parallel MSYS2 process
startup is common:
- https://github.com/git-for-windows/git/issues/493
- https://github.com/git-for-windows/git/issues/4076
- https://github.com/git-for-windows/git/issues/3870
- https://github.com/anthropics/claude-code/issues/30165

## Changes

**mount.cc: Add `MOUNT_OVERRIDE` to `create_root_entry()`**

This is the direct fix. If `/` already exists in the mount table when
`create_root_entry()` runs, `add_item()` now overwrites it instead of
returning `EPERM`. This makes the root mount initialization idempotent.

**shared.cc: Use `created` flag from `open_shared()` in `user_info::create()`**

`open_shared()` already tracks whether the shared memory mapping was
newly created or already existed (via `CreateFileMappingW` /
`ERROR_ALREADY_EXISTS`). Currently `user_info::create()` calls the
overload that discards this flag.

This change switches to the overload that returns it, and calls
`initialize()` when `created` is true. This provides an earlier
initialization path for the process that created the shared memory,
before `dll_crt0_1()` reaches `initialize()` via the spinlock.

Note: this does not replace the existing spinlock coordination in
`initialize()`. The normal cold-start path through `dll_crt0_1()` is
unchanged. This is a supplementary improvement, not the primary fix.

## Testing

Built patched `msys-2.0.dll` from Cygwin 3.6.7 with MSYS2 patches.
Tested on Windows 11 Enterprise with parallel bash spawning from
Claude Code (which fires multiple hook processes per tool invocation).
The crash was occurring consistently before the patch and has not
recurred after installing it.

### DIFF
--- a/winsup/cygwin/mm/shared.cc
+++ b/winsup/cygwin/mm/shared.cc
@@ -221,6 +221,7 @@ void
 user_info::create (bool reinit)
 {
   WCHAR name[UNLEN + 1] = L""; /* Large enough for SID */
+  bool created = false;
 
   if (reinit)
     {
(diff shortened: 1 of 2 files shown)


ilude:
Happy to submit this upstream to cygwin-patches if that's the preferred path. Opened here first since this is where I built and tested against, but understand if this belongs upstream given both files are core Cygwin code.

dscho:
@ilude sorry for the long delay; I really recommend submitting this upstream, as it affects Cygwin (and thus reviewing requires the expertise on the [cygwin-patches mailing list](https://inbox.sourceware.org/cygwin-patches/)).

PR #334, opened 2026-04-07
msys-2.0.dl to cygwin1.dll switch patches

lygstate:
This MR reorder the patches, so we can transition from msys-2.0.dl to cygwin1.dll.

### First, we revert the patches that rename cygwin1.dll to msys-2.0.dll

* Revert "strace.cc: Don't set MSYS=noglob"

* Revert "Rename dll from cygwin to msys"

* Revert "Fix msys library name in import libraries"

* Revert "uname: allow setting the system name to CYGWIN"

* Revert "Add functionality for changing OS name via MSYSTEM environment variables."

NOTE: these **reverts** can be **dropped** in the next rebase

### Then add patches that support for transition msys-2.0.dl to cygwin1.dll.

* Add functionality for changing OS name via MSYSTEM environment variables.

* Support for both cygwin and msys2

* Support for both MSYS and CYGWIN environment variable

* strace.cc: Don't set CYGWIN=noglob

* Convert to use msys'style named pipe

### The last two patch is to keep the orignal ` Rename dll from cygwin to msys`

* Rename dll from cygwin to msys

* Fix msys library name in import libraries

With the last two patch, the built dll  will be `msys-2.0.dll`.
Without the last two patch, the built dll  will be `cygwin1.dll`.

So we can bootstrap msys-2.0.dll to cygwin1.dll in `MSYS2-packages` in a single code base.


dscho:
> This MR reorder the patches, so we can transition from msys-2.0.dl to cygwin1.dll.

I am unaware of such a plan.

I mean, I can imagine that there is a ton of benefit to reduce the difference between the MSYS2 runtime and the Cygwin runtime to the bare minimum, to the point where even the path translation (which is probably _the_ most distinguishing feature of the MSYS2 runtime) would be upstreamed to the Cygwin runtime as an opt-in, e.g. via a flag in `/etc/nsswitch.conf`. That would generate synergies between the Cygwin and the MSYS2 project, making it even easier to help each other out in meaningful ways. I'm all for going this direction.

At the same time, I am convinced that such a plan can only succeed by first building consensus about the direction, and I haven't seen such a discussion. Have I missed it?

lygstate:
> > This MR reorder the patches, so we can transition from msys-2.0.dl to cygwin1.dll.
> 
> I am unaware of such a plan.
> 
> I mean, I can imagine that there is a ton of benefit to reduce the difference between the MSYS2 runtime and the Cygwin runtime to the bare minimum, to the point where even the path translation (which is probably _the_ most distinguishing feature of the MSYS2 runtime) would be upstreamed to the Cygwin runtime as an opt-in, e.g. via a flag in `/etc/nsswitch.conf`. That would generate synergies between the Cygwin and the MSYS2 project, making it even easier to help each other out in meaningful ways. I'm all for going this direction.
> 
> At the same time, I am convinced that such a plan can only succeed by first building consensus about the direction, and I haven't seen such a discussion. Have I missed it?

This is my local draft work that can transition from transition msys-2.0.dl to cygwin1.dll, post here for discussion.

This is just barelly bootstrap msys-2.0.dll into cygwin1.dll, other things I am not considered yet.

The most important part of msys2 I think is `pacman`, othere things(the runtime differencies) is minimal now.

And indeed pacman not working in cygwin at all.

winsup/cygwin/hookapi.cc
```diff
@@ -378,13 +378,10 @@ hook_or_detect_cygwin (const char *name, const void *fn, WORD& subsys, HANDLE h)
...
-      if (!ascii_strcasematch (rva (PSTR, map ?: (char *) hm, pd->Name - delta),
-#ifdef __MSYS__
-			       "msys-2.0.dll"))
-#else
-			       "cygwin1.dll"))
-#endif
-	continue;
+      char *name_to_compare = rva (PSTR, map ?: (char *) hm, pd->Name - delta);
+      bool not_msys2 = !ascii_strcasematch (name_to_compare, "msys-2.0.dll");
+      bool not_cygwin = !ascii_strcasematch (name_to_compare, "cygwin1.dll");
```
dscho, line 383: Up until this point, I see the keying factor to be the presence of the `MSYS` and the `CYGWIN` environment variable. It is not quite clear what this MR wants to do if both are set, but it is clear that it wants to trigger the MSYS2 vs Cygwin code paths depending on the presence of either environment variable.

This hunk, however, tells a very different story (and one that I could understand much better): whether to run in  MSYS2 or in Cygwin mode is keyed by the name of the DLL.

It sounds a bit fraught with peril to have _both_ mechanisms. It would probably be a lot more desirable to key only on one of the two, not on both.

winsup/utils/mingw/strace.cc
```diff
@@ -356,36 +356,21 @@ create_child (char **argv)
...
     space = cygwin_env = "";
-  char *newenv = (char *) malloc (sizeof ("MSYS=noglob")
+  char *newenv = (char *) malloc (sizeof ("CYGWIN=noglob")
 				  + strlen (space) + strlen (cygwin_env));
-  sprintf (newenv, "MSYS=noglob%s%s", space, cygwin_env);
+  sprintf (newenv, "CYGWIN=noglob%s%s", space, cygwin_env);
   _putenv (newenv);
-*/
+#endif
+
```
dscho, line 373: That's a lot of dead code. It would make more sense to investigate properly what the original commit tried to do, why it was wrong, and then to drop that part altogether. But that should most likely come as a _separate_ Merge Request.

winsup/cygwin/hookapi.cc
lygstate: This is a temporary patch that bootstrap from "msys-2.0.dll" to "cygwin1.dll", once the bootstrap finished. we can then revert this patch directly.

winsup/utils/mingw/strace.cc
lygstate: I do this for reduce `MSYS` presence in code-base, as it's commted code, so I use #if 0, as this can be accepted, I'll create separte MR for it.

lygstate:
> I mean, I can imagine that there is a ton of benefit to reduce the difference between the MSYS2 runtime and the Cygwin runtime to the bare minimum, to the point where even the path translation (which is probably the most distinguishing feature of the MSYS2 runtime) would be upstreamed to the Cygwin runtime as an opt-in, e.g. via a flag in /etc/nsswitch.conf. That would generate synergies between the Cygwin and the MSYS2 project, making it even easier to help each other out in meaningful ways. I'm all for going this direction.

The main target of this MR is rename rename `msys-2.0.dll` back to `cygwin1.dll`, and convert `msys-` prefix ino `cyg`, so that the whole `MYS2-packages` can remove many patches, the upstream patches of msys2-runtime into cygwin did not conflict with this MR, as this MR did not change the runtime much.

PR #340, opened 2026-05-18
cygcheck: remove an unused variable causing a build error with GCC 16 + zstd CI fix

lazka:
Since gcc 16 this triggers -Werror=unused-but-set-variable= Remove the unused variable.

```c++
../../../.././winsup/utils/mingw/cygcheck.cc: In function 'void dump_sysinfo()':
../../../.././winsup/utils/mingw/cygcheck.cc:1709:11: error: variable 'count_path_items' set but not used [-Werror=unused-but-set-variable=]
 1709 |       int count_path_items = 0;
      |           ^~~~~~~~~~~~~~~~
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,7 +18,7 @@ jobs:
         with:
           msystem: MSYS
           update: true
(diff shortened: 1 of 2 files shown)


lazka:
The new zstd dependency for the dumper is due to https://github.com/msys2/MSYS2-packages/pull/6240 I guess

PR #342, opened 2026-06-01
Fix #176. Use the CRT's signal handling on native Win32 processes.

### LINKED ISSUE #176: kill() / signal() doesn't really work.
Currently MSYS isn't able to send signals to ordinary Win32 programs. A long time ago I wrote a patch for MSYS1 to accomplish this. https://sourceforge.net/p/mingw/mailman/message/10716023/

It would need refreshing to handle 64bit and 32bit binaries but otherwise should still work. Is there any interest in supporting this?

hyc:
This allows the 7 signals that Microsoft's C runtime supports to be used, instead of only Ctrl-C/SIGINT.

This is just a slightly refreshed version of the patch I wrote for MSYS1 twenty years ago. https://sourceforge.net/p/mingw/bugs/1783/

It only looks for ucrtbase.dll and msvcrt.dll now, and it doesn't worry about 32bit vs 64bit processes since 32bit Windows is long obsolete.

### DIFF
--- a/winsup/cygwin/exceptions.cc
+++ b/winsup/cygwin/exceptions.cc
@@ -1524,6 +1524,7 @@ sigpacket::process ()
   bool issig_wait = false;
   struct sigaction& thissig = global_sigs[si.si_signo];
   void *handler = have_execed ? NULL : (void *) thissig.sa_handler;
+  extern char *pExecedHook;
 
   threadlist_t *tl_entry = NULL;
   _cygtls *tls = NULL;
@@ -1627,6 +1628,54 @@ sigpacket::process ()
     goto exit_sig;
   if (si.si_signo == SIGSTOP)
     goto stop;
+
+/* This is the value the MSVC RT uses for SIGBREAK. MSVC doesn't define
+ * SIGQUIT so we turn it into SIGBREAK, as that seems to be the intent
+ * of the signal. MSVC also sends SIGBREAK to NT Services in response
+ * to CTRL_LOGOFF_EVENTs, which we would consider SIGHUP. Since MSVC
+ * doesn't handle SIGHUP we translate SIGHUP to SIGBREAK as well.
+ */
+#ifndef SIGBREAK
+#define SIGBREAK 21
+#endif
+  /* both of these may be used for SIGABRT */
+#define SIGABRT_22	22
+#define SIGABRT_6	6
+
+  if (pExecedHook)
+    {
+      HANDLE hRemThread;
+      SIZE_T nbytes;
+      int rsig = si.si_signo;
+      switch(rsig)
+	{
+	  case SIGQUIT:
+	  case SIGHUP:
+	    rsig = SIGBREAK;
+	    fallthrough;
+	  case SIGINT:
+	  case SIGTERM:
+	  case SIGBREAK:
+	  case SIGABRT_22:
+	  case SIGABRT_6:
+	  case SIGFPE:
+	  case SIGILL:
+	  case SIGSEGV:
+	    WriteProcessMemory(ch_spawn,pExecedHook,&rsig,sizeof(rsig),&nbytes);
+	    hRemThread = CreateRemoteThread(ch_spawn,NULL,0,
+		      (LPTHREAD_START_ROUTINE)(pExecedHook+sizeof(remote_info1)),
+		      pExecedHook, 0, NULL);
+	    if (hRemThread)
+	      {
+	        WaitForSingleObject(hRemThread,INFINITE);
+		CloseHandle(hRemThread);
+		goto done;
+	      }
+	    fallthrough;
+	  default:
+	    break;
+	}
+    }
 
   if (handler == (void *) SIG_DFL)
     {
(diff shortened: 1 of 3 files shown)


hyc:
Could also just do this in a standalone kill program if you wanted that instead.

[wkill.c](https://github.com/user-attachments/files/28449886/wkill.c)

dscho:
> It only looks for ucrtbase.dll and msvcrt.dll now, and it doesn't worry about 32bit vs 64bit processes since 32bit Windows is long obsolete.

This is a non-sequitur. While the 32-bit _Windows_ is long obsolete, indeed, 32-bit Windows _processes_ are not.

> ```c
> 	    WriteProcessMemory(ch_spawn,pExecedHook,&rsig,sizeof(rsig),&nbytes);
> ```

This is a dark pattern, and is likely to land a MSYS2 runtime that uses that pattern _much_ more likely in quarantine than before, and rightfully so. I know that we [inject a remote thread](https://github.com/msys2/msys2-runtime/blob/01d6c708f9221334d18ab332621b6d87eb12d37e/winsup/utils/mingw/getprocaddr.c#L38) to send signals. The difference is that we go out of our way _not to inject executable code_, but to call well-known entry points (`CtrlRoutine`/`ExitProcess`) instead.

Seeing that there is already this pattern, I have to wonder why you didn't simply choose to use the very same method: [`raise()`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/raise?view=msvc-170) accepts an `int`, which is compatible with the type of the [`lpParameter`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createremotethread#:~:text=lpParameter,-A) we can pass to the remote thread start routine. So: why go out of your way to ignore the existing code pattern and invent a new, incompatible one instead of imitating it (passing the address of `raise()` as the `lpStartAddress`)?

hyc:
> Seeing that there is already this pattern, I have to wonder why you didn't simply choose to use the very same method: [raise()](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/raise?view=msvc-170) accepts an int, which is compatible with the type of the [lpParameter](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createremotethread#:~:text=lpParameter,-A) we can pass to the remote thread start routine. So: why go out of your way to ignore the existing code pattern and invent a new, incompatible one instead of imitating it (passing the address of raise() as the lpStartAddress)?

That's not correct. The parameter to a ThreadProc is a pointer. As raise() takes an int, and pointer and int are different sizes, you can't just use raise() itself as the start address.

https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms686736(v=vs.85)

Also, you can't rely on just knowing the address of raise() in the target process; the only way to know it for certain is to use GetProcAddress in the target process. Especially if the target process is using a CRT DLL that the calling process doesn't use.

As for the other comment, I can still look into 32bit support. Just makes the build a little more complicated.

dscho:
> > Seeing that there is already this pattern, I have to wonder why you didn't simply choose to use the very same method: [raise()](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/raise?view=msvc-170) accepts an int, which is compatible with the type of the [lpParameter](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createremotethread#:~:text=lpParameter,-A) we can pass to the remote thread start routine. So: why go out of your way to ignore the existing code pattern and invent a new, incompatible one instead of imitating it (passing the address of raise() as the lpStartAddress)?
> 
> That's not correct. The parameter to a ThreadProc is a pointer. As raise() takes an int, and pointer and int are different sizes, you can't just use raise() itself as the start address.

In this context, the pointer type is always at least as wide as the int type. So there needs to be at least a whiff of some thought having been put behind this change _intentionally_ disagreeing with the existing code pattern.

However, the most worrisome part about this PR is the continued insistence on a coding pattern that is _prone_ to land the MSYS2 runtime into pretty much every anti-malware's quarantine, even the ones which are graciously absent from the list of common false positive generators.

You are not addressing this issue. This code change _will_ make the MSYS2 runtime look like malware. It _will_ cause more trouble than we want to buy into.

rimrul:
> While the 32-bit Windows is long obsolete, indeed, 32-bit Windows processes are not.

That's not even all. On Windows 11 On ARM  we need to consider x86, x86-64, ARM32 (until 24H2), ARM64 and ARM64EC processes.

Edit for clarity: by "need to consider" I don't  necessarily mean that we need to use their CRTs signal handling, just that we need to be aware we can encounter them and should have some fallback for when we do.

hyc:
> In this context, the pointer type is always at least as wide as the int type. So there needs to be at least a whiff of some thought having been put behind this change intentionally disagreeing with the existing code pattern.

Don't be an ass. Of course this was thought about. And again, you can't reliably know the address of raise() in the target process without sending custom code over to find it. And this pattern isn't "new" - it was accepted in MSYS1 20 years ago.

dscho:
> Don't be an ass.

I bow out of this conversation. If you cannot converse with me respectfully, you cannot develop code with me.

If others with write permission to this repository want to continue discussing this change, I won't stop them. Otherwise I will close this PR.

hyc:
>  If you cannot converse with me respectfully, you cannot develop code with me.

If you question whether thought was put into a contribution, it is you who needs to be more respectful.

PR #343, opened 2026-06-11
Cygwin: align testsuite tmp directory with runtime path resolution

aswin-mcw:
## Summary

While investigating Cygwin testsuite failures in MSYS2, I found that the installation root was being resolved one directory too far back in my build layout.

In `init_cygheap::init_installation_root()` (`winsup/cygwin/mm/cygheap.cc`), the `__MSYS__` logic strips three directory levels in total when determining the installation root. For my test environment, this resulted in incorrect path resolution.

The testsuite was also creating temporary files under `${builddir}/testinst/tmp`. Changing it to use `${builddir}/tmp` aligns the test layout with the runtime's expected path resolution.

```diff
-$(MKDIR_P) ${builddir}/testinst/tmp
+$(MKDIR_P) ${builddir}/tmp

-rm -rf ${builddir}/testinst/tmp
+rm -rf ${builddir}/tmp
```

## Results

After making this change, the winsup testsuite results improved significantly in my environment:

* Before: ~148 passing tests out of 282
* After: ~258 passing tests out of  282

This suggests that a large number of failures were caused by path resolution and test layout issues rather than individual runtime failures.

### DIFF
--- a/winsup/cygwin/Makefile.am
+++ b/winsup/cygwin/Makefile.am
@@ -620,8 +620,8 @@ $(NEW_DLL_NAME): $(LDSCRIPT) libdll.a $(VERSION_OFILES) $(LIBSERVER)\
 	$(newlib_build)/libm.a \
 	$(newlib_build)/libc.a \
 	-lgcc -lkernel32 -lntdll -Wl,-Map,msys.map
-	@$(MKDIR_P) ${target_builddir}/winsup/testsuite/testinst/bin/
-	$(AM_V_at)$(INSTALL_PROGRAM) $(NEW_DLL_NAME) ${target_builddir}/winsup/testsuite/testinst/bin/$(DLL_NAME)
+	@$(MKDIR_P) ${target_builddir}/winsup/testsuite/testinst/usr/bin/
+	$(AM_V_at)$(INSTALL_PROGRAM) $(NEW_DLL_NAME) ${target_builddir}/winsup/testsuite/testinst/usr/bin/$(DLL_NAME)
 
 # msys-2.0 import library
 toolopts=--cpu=@target_cpu@ --ar=@AR@ --as=@AS@ --nm=@NM@ --objcopy=@OBJCOPY@
--- a/winsup/testsuite/Makefile.am
+++ b/winsup/testsuite/Makefile.am
@@ -343,7 +343,7 @@ XFAIL_TESTS = \
 # cygrun.sh test-runner script, and variables used by it:
 LOG_COMPILER = $(srcdir)/cygrun.sh
 
-export runtime_root=$(abs_builddir)/testinst/bin
+export runtime_root=$(abs_builddir)/testinst/usr/bin
 export cygrun=$(builddir)/mingw/cygrun
 
 # Set up things in the Cygwin 'installation' at testsuite/testinst/ to provide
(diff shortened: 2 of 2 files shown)


dscho:
I have to admit that I am a bit puzzled by this. As per [this comment](https://github.com/aswin-mcw/msys2-runtime/blob/4831fa6d6c3fc8c117d5684c8ef5be8974d500d5/winsup/cygwin/mm/cygheap.cc#L223-L227), the MSYS2 runtime is expected to have one extra level relative to the Cygwin runtime, not one less: The `cygwin1.dll` is supposed to be in `/bin/`, whereas the MSYS2 runtime wants to live in `/usr/bin/`.

aswin-mcw:
> I have to admit that I am a bit puzzled by this. As per [this comment](https://github.com/aswin-mcw/msys2-runtime/blob/4831fa6d6c3fc8c117d5684c8ef5be8974d500d5/winsup/cygwin/mm/cygheap.cc#L223-L227), the MSYS2 runtime is expected to have one extra level relative to the Cygwin runtime, not one less: The `cygwin1.dll` is supposed to be in `/bin/`, whereas the MSYS2 runtime wants to live in `/usr/bin/`.

Agreed — the "Back two folders to get root as we have all stuff in usr subfolder" logic you linked is exactly right, and it's not what I'm touching. MSYS2 keeps the runtime one level deeper (/usr/bin/msys-2.0.dll vs Cygwin's /bin/cygwin1.dll), so init_installation_root() correctly strips that extra usr level.

The bug was that the testsuite didn't reproduce that layout: it staged the test DLL and the sh/sleep/ls helpers under testinst/bin (the Cygwin 2-level layout, no usr). So the runtime's 3-level strip had no usr to consume and overshot the installation root to the parent of testinst (${builddir}) — which is why /tmp resolved to ${builddir}/tmp instead of testinst/tmp. My earlier testinst/tmp → ${builddir}/tmp change was just chasing that overshot root, so you're right that it was the wrong direction.

I've replaced it with a layout fix so the test tree mirrors a real MSYS2 install:

winsup/cygwin/Makefile.am — stage the test DLL under testinst/usr/bin (was testinst/bin).
winsup/testsuite/Makefile.am — point runtime_root and the sh/sleep/ls helpers at testinst/usr/bin; leave testinst/tmp unchanged.
Now the strip lands on testinst, /bin and /usr/bin both map to testinst/usr/bin, and /tmp stays testinst/tmp — no path hack, and the test env finally has the /usr/bin level the runtime assumes. With the matching CI staging change in #344, the winsup testsuite goes from ~148/282 to 264/282 .

dscho:
@aswin-mcw I opened a PR to fix [that (new) build failure](https://github.com/msys2/msys2-runtime/actions/runs/30452512302/job/90577775084) you got: #355. I merged that, rebased your branch on top and changed the target branch to the now-current default branch.

aswin-mcw:
> @aswin-mcw I opened a PR to fix [that (new) build failure](https://github.com/msys2/msys2-runtime/actions/runs/30452512302/job/90577775084) you got: #355. I merged that, rebased your branch on top and changed the target branch to the now-current default branch.

Thanks a lot, @dscho — appreciate you fixing the build failure in #355, merging it, and rebasing/retargeting this branch for me. I've pulled the rebased branch and confirmed it builds and the winsup testsuite still passes at the expected rate (264/282) against the new default branch, and CI here looks green now. Let me know if there's anything else you'd like me to adjust before merge.

dscho:
@aswin-mcw I'd like to understand those failures first, to see whether we _can_ fix them or not.

The file mode related ones most likely need a `/tmp/` that is mounted with `acl` (instead of the implicit `noacl` as of MSYS2's https://github.com/msys2/msys2-runtime/commit/67e0663d5a6b3ba59a672c430dcab9152bd19d87#diff-0952970e73771f7d9057ddbdd172956d54bc76b4dc3ac3f7c98cba1b0a547671R44-R553), for example, and the `msgtest` one only fails in that testinst for me, when I run it interactively (after starting a `cygserver` in the same MSYS2 environment), it succeeds.

dscho:
In other words, I'd like to proceed at https://github.com/msys2/msys2-runtime/pull/344#issuecomment-5125532839 (i.e. have a single PR that fixes the Cygwin testsuite for MSYS2 and runs it as part of CI).

Issue #135, opened 2023-01-15
Run unit tests in CI


lazka:
It would be nice if we would run cygwin tests in CI, so we can sleep better.

Running unit tests in 3.4 is currently broken in general (this should fix the basics: https://cygwin.com/git/?p=newlib-cygwin.git;a=commit;h=4e7817498efc8ad74b7266d034abc5c4e45de44e)

There have been a bunch of fixes in cygwin master to improve the tests, so we could either backport them, or wait for 3.5.

dscho:
I agree it would be nice. So I gave it a try, and I need these changes thus far:

```diff
diff --git a/winsup/testsuite/Makefile.am b/winsup/testsuite/Makefile.am
index 2289556682..472263cd87 100644
--- a/winsup/testsuite/Makefile.am
+++ b/winsup/testsuite/Makefile.am
@@ -318,7 +318,7 @@ check_PROGRAMS = \
 # flags for linking against the just built implib
 # TODO: use -nostdinc and to-be-installed headers as well?
 LDFLAGS_FOR_TESTDLL = -nodefaultlibs
-LDADD_FOR_TESTDLL = $(builddir)/../cygwin/libcygwin.a -lgcc -lkernel32 -luser32
+LDADD_FOR_TESTDLL = $(builddir)/../cygwin/libmsys-2.0.a -lgcc -lkernel32 -luser32
 
 # flags for test executables
 AM_CPPFLAGS = -I$(srcdir)/libltp/include
diff --git a/winsup/testsuite/cygrun.sh b/winsup/testsuite/cygrun.sh
index bf1d5cc6b5..29d107b22f 100755
--- a/winsup/testsuite/cygrun.sh
+++ b/winsup/testsuite/cygrun.sh
@@ -11,7 +11,7 @@ export PATH="$runtime_root:${PATH}"
 if [ "$1" = "./mingw/cygload" ]
 then
     windows_runtime_root=$(cygpath -m $runtime_root)
-    $cygrun "$exe -v -cygwin $windows_runtime_root/cygwin1.dll"
+    MSYS_NO_PATHCONV=1 $cygrun "$exe -v -cygwin $windows_runtime_root/msys-2.0.dll"
 else
-    cygdrop $cygrun $exe
+    $cygrun $exe
 fi
```

However, even with this, most tests don't seem to pass (this is running `make -j15 -C x86_64-pc-msys/winsup/testsuite/ check-TESTS AM_COLOR_TESTS=always`):

[...]

PR #344, opened 2026-06-15
github-actions: add Windows CI for Cygwin testsuite

aswin-mcw:
## Summary

Run Cygwin's testsuite against the MSYS2 runtime in CI on Windows, and add the
fixes and (justified) XFAIL annotations needed to bring the suite up cleanly on
x86_64. The testsuite is wired in as an additional job on the existing
`build.yaml` workflow (reusing its build output), following the commit
structure suggested in review.

## Implementation

The Cygwin testsuite runs as a new `cygwin-tests` job in
`.github/workflows/build.yaml` instead of a standalone workflow, so it reuses
the existing `build` job rather than duplicating the build steps:

- The `build` job additionally compiles the test programs (`make check_programs`)
  and uploads the built worktree as an artifact.
- The `cygwin-tests` job downloads that worktree, starts `cygserver`, and runs
  `make check` with `MSYS=winsymlinks:nativestrict`.
- Test logs (`*.log` / `*.trs`) are uploaded for debugging/regression analysis.

The freshly built `msys-2.0.dll` is staged under `testinst/usr/bin` by the
runtime's own install rules, so the tests exercise the just-built runtime.

## Commit structure

This follows the structure requested in review: the individual fixes are
`fixup!` commits targeting the correct topic commits (so they squash into the
right place on the next rebase onto a newer Cygwin), line endings are enforced
via `.gitattributes`, and the CI change is a single commit on `build.yaml`. The
only standalone additions are the two XFAIL commits (each with a full rationale)
and small CI-plumbing fixes.

## Fixes included

- **UNC-style path fix** — when cwd or a `PATH` entry is `/`, path construction
  produced `//<cmd>`, which Windows/MSYS2 interprets as a UNC network path,
  breaking command resolution from the install root. Fixed by treating `/` as an
  empty prefix. Fixes `fcntl07`, `fcntl07b`, `cancel11`.
- **busybox setup fix** — the syscall test needs `sh`/`sleep`/`ls` from busybox,
  but the previous setup pointed at a non-existent path
  (`/usr/libexec/busybox/bin`). Fixed to copy from `/usr/bin`, staged under the
  `/usr/bin` layout, with `busybox` added to the CI packages.
- **Test DLL linkage** — test DLLs link against `libmsys-2.0`.
- **cygserver startup race** — `cygserver` is started asynchronously; the SysV
  IPC tests (`semtest`/`shmtest`/`msgtest`) would race its startup and fail
  intermittently. The job now waits for `cygserver` to be up before running the
  suite.

## Tests marked XFAIL (with rationale)

**Permission tests requiring ACL-backed mounts** — `access01`, `access05`,
`chmod01`, `stat02`, `umask03`, and the EACCES case of `symlink03`. MSYS2 mounts
with `noacl` by default, so POSIX permission bits are not backed by Windows ACLs
and `chmod()`/`umask()` cannot enforce them; Cygwin mounts with ACLs by default,
which is why these pass there. Enabling `acl` was tried and is not viable: it
makes this group pass but regresses `stat01`/`fstat02`/`fstat04`, doesn't fix
`umask03`, and runs a non-default configuration — so it only trades one set of
failures for another. (The full reasoning is in the commit message.)

**Symlink tests needing non-native semantics** — `symlink01`, `symlink05`. With
`MSYS=winsymlinks:nativestrict` most symlink tests now pass (`lstat02`,
`readlink01/02`, `symlink02/04`). These two require creating a symlink to a
non-existing ("dangling") target, which native NTFS symlinks cannot represent,
so they fail in setup under `nativestrict`.

## Known failure (not masked)

- **`mingw/cygload`** — left as a real FAIL rather than XFAIL'd. It is a native
  (non-MSYS2) program that `LoadLibrary()`s the runtime DLL, and it fails with
  error 126 because the absolute DLL path passed by `cygrun.sh` is mangled by
  MSYS2's automatic argument path conversion (`D:/a/...` becomes the invalid
  `D;A:\...`). This is a test-harness path-passing bug, not a runtime defect;
  a clean fix is still being worked out.

## Current results (x86_64)

```
TOTAL: 282 | PASS: 269 | XFAIL: 12 | FAIL: 1 (mingw/cygload) | SKIP/XPASS/ERROR: 0
```

Up from ~151 passing. This establishes an automated baseline for running
Cygwin's testsuite against the MSYS2 runtime on x64 and a foundation for
tracking regressions.


.github/workflows/cygwin.yml
```diff
@@ -106,66 +106,74 @@ jobs:
...
+    - name: Build
+      shell: msys2 {0}
+      run: |
+        (cd winsup && ./autogen.sh)
+        mkdir build && cd build
+        ../configure \
+          --prefix=/usr \
+          --disable-doc \
+          --disable-dumper
+        make -j$(nproc) MINGW_CXXFLAGS="-Wno-error=unused-but-set-variable"
```
dscho: That merely duplicates what [the `build` workflow already does](https://github.com/msys2/msys2-runtime/blob/8fbd9808447ee78ed485deead9b79cd1e40c07b7/.github/workflows/build.yaml#L23-L28). Can't we simply add running the Cygwin test suite as an additional job to the [`build.yaml` definition](https://github.com/msys2/msys2-runtime/blob/msys2-3.6.10/.github/workflows/build.yaml)?

winsup/testsuite/winsup.api/ltp/access01.c
```diff
@@ -148,7 +148,6 @@ int main(int ac, char **av)
         int experrno;
     } Test_cases[] = {
         { Fname, F_OK,  "F_OK", 0 },
-        { Fname, X_OK,  "X_OK", 0 },
```
dscho, line 151: I'm not quite certain why this should fail with MSYS2 when it succeeds with Cygwin. After all, Cygwin has to contend with the same ACL vs POSIX permissions incompatibility.

I'm not saying that this patch is wrong, what I'm saying is that the commit message _needs_ to provide a good rationale for this. I like to point people to the guidance in https://github.blog/2022-06-30-write-better-commits-build-better-projects/ to improve commit messages, in particular with a strong focus on this part:

  | What you’re doing | Why you’re doing it
-- | -- | --
High-level (strategic) | **Intent** (what does this accomplish?) | **Context** (why does the code do what it does now?)
Low-level (tactical) | **Implementation** (what did you do to accomplish your goal?) | **Justification** (why is this change being made?)

winsup/testsuite/Makefile.am
```diff
@@ -341,6 +341,13 @@ XFAIL_TESTS = \
 	winsup.api/ltp/umask03 \
 	winsup.api/ltp/stat02 \
 	winsup.api/ltp/chmod01 \
+	winsup.api/ltp/lstat02 \
+	winsup.api/ltp/readlink01 \
+	winsup.api/ltp/readlink02 \
+	winsup.api/ltp/symlink01 \
+	winsup.api/ltp/symlink03 \
+	winsup.api/ltp/symlink04 \
+	winsup.api/ltp/symlink05 \
```
dscho: Would these tests not succeed with `MSYS=winsymlinks:nativestrict`?

winsup/testsuite/Makefile.am
```diff
@@ -375,9 +375,10 @@ export cygrun=$(builddir)/mingw/cygrun
 
 check-local:
 	$(MKDIR_P) ${builddir}/testinst/tmp
-	cd ${builddir}/testinst/bin && cp /usr/libexec/busybox/bin/busybox.exe sh.exe
-	cd ${builddir}/testinst/bin && cp /usr/libexec/busybox/bin/busybox.exe sleep.exe
-	cd ${builddir}/testinst/bin && cp /usr/libexec/busybox/bin/busybox.exe ls.exe
+	$(MKDIR_P) ${builddir}/usr/bin
+	cd ${builddir}/usr/bin && cp /usr/bin/busybox.exe sh.exe
+	cd ${builddir}/usr/bin && cp /usr/bin/busybox.exe sleep.exe
+	cd ${builddir}/usr/bin && cp /usr/bin/busybox.exe ls.exe
```
dscho: I agree with this change, but it should be marked as `fixup! - Move root to /usr. - Change sorting mount points. - By default mount without ACLs. - Can read /etc/fstab with short mount point format.` so that it gets squashed into the correct commit upon the next rebase of the MSYS2 runtime to the next Cygwin version.

dscho (commented):
Thank you for working on this! I'm still not quite certain that those failures should be expected. Ideally we would really try hard to let them all succeed. I vaguely remember running the entire test suite at some point and running into tremendous problems before running `cygserver`, but I lost my notes on that experiment. I guess I'll try to re-run this on my side ;-)

dscho:
> I vaguely remember running the entire test suite at some point and running into tremendous problems before running `cygserver`, but I lost my notes on that experiment.

I found them again! https://github.com/msys2/msys2-runtime/issues/135#issuecomment-1967438362

dscho:
For the record: https://github.com/msys2/msys2-runtime/compare/dscho:also-run-cygwin-testsuite-in-ci demonstrates the commit structure that I would strongly prefer.

aswin-mcw:
> Thank you for working on this! I'm still not quite certain that those failures should be expected. Ideally we would really try hard to let them all succeed. I vaguely remember running the entire test suite at some point and running into tremendous problems before running `cygserver`, but I lost my notes on that experiment. I guess I'll try to re-run this on my side ;-)

Thanks! I took "try hard to let them succeed" seriously and reworked this substantially. With cygserver started before make check and MSYS=winsymlinks:nativestrict, the suite now runs at 269 PASS / 12 XFAIL / 1 FAIL (up from ~151 passing).

You were right that cygserver is essential — the SysV IPC tests (semtest/shmtest/msgtest) fail without it. I also hit a variant of your "tremendous problems": cygserver starts asynchronously and the IPC tests were racing its startup and failing intermittently, so the job now waits for cygserver to be up before running the suite.

Rather than blanket-XFAIL the rest, I investigated each one and only kept XFAILs that are genuine platform limitations, each justified in its commit message (see the per-thread replies). The one honest exception is mingw/cygload, which I've deliberately left as a real FAIL rather than mask it — it's a test-harness path-mangling bug (the DLL path is corrupted to D;A:\… before LoadLibrary), not a runtime defect; I'm still working on a clean fix.

.github/workflows/cygwin.yml
aswin-mcw: > That merely duplicates what [the `build` workflow already does](https://github.com/msys2/msys2-runtime/blob/8fbd9808447ee78ed485deead9b79cd1e40c07b7/.github/workflows/build.yaml#L23-L28). Can't we simply add running the Cygwin test suite as an additional job to the [`build.yaml` definition](https://github.com/msys2/msys2-runtime/blob/msys2-3.6.10/.github/workflows/build.yaml)?

Agreed — done. I removed the standalone cygwin.yml and added the testsuite as a cygwin-tests job in build.yaml that reuses the existing build job's output (it uploads the built worktree as an artifact; the test job downloads it and runs make check) instead of rebuilding. No duplication now.

winsup/testsuite/winsup.api/ltp/access01.c
aswin-mcw: You're right that "MSYS2 doesn't replicate chmod semantics" isn't a real rationale. The actual cause is MSYS2's default noacl mount: permission bits aren't backed by Windows ACLs, so chmod()/umask() can't enforce them (the mode comes from the DOS read-only attribute only). Cygwin mounts with ACLs by default, which is why the same tests pass there.

I verified this empirically instead of assuming: mounting the test tree with acl does make access01/05, chmod01, stat02 pass — but it regresses stat01/fstat02/fstat04 (they then read ACL-derived owner/mode values the tests don't expect), doesn't fix umask03/symlink03, and runs a config MSYS2 doesn't use by default. So acl only trades one set of failures for another. (An earlier run looked like acl broke the cygserver IPC tests — that was actually the cygserver startup race, not ACLs.)

This is now captured in the commit message (Cygwin: testsuite: XFAIL permission tests that fail under MSYS2's noacl default), following the intent/context/implementation/justification structure you linked.

winsup/testsuite/Makefile.am
aswin-mcw: "
Good call — with MSYS=winsymlinks:nativestrict (which the job now sets) most of them pass, so I dropped the blanket symlink XFAILs: lstat02, readlink01/02, symlink02/04 now pass. Three remain, for specific reasons:

symlink05 — creates a symlink to a non-existing (dangling) target and expects success; a native NTFS symlink must record the target type at creation, so nativestrict can't create a dangling link.
symlink01 — its setup creates symlinks to both real and non-existing ((null)) targets and verifies them with lstat(); the dangling ones can't be created under nativestrict, so it aborts in setup ("Test Case Declaration Error"). Its ELOOP / over-long-pathname case actually passes.
symlink03 — its failing case expects EACCES when the target directory isn't writable, which is the same noacl permission-enforcement gap as the chmod group.
The two link modes winsymlinks:lnk/:sys would let these pass but regress the many tests relying on native symlink semantics, so nativestrict stays the default. Both are spelled out in the commit messages.

winsup/testsuite/Makefile.am
aswin-mcw: Done — this is now a fixup! - Move root to /usr. … commit so it squashes into that topic commit on the next rebase onto the newer Cygwin version.

aswin-mcw:
> > I vaguely remember running the entire test suite at some point and running into tremendous problems before running `cygserver`, but I lost my notes on that experiment.
> 
> I found them again! [#135 (comment)](https://github.com/msys2/msys2-runtime/issues/135#issuecomment-1967438362)

Thanks for digging those up — that matches exactly what we saw. The cygwin-tests job starts cygserver before make check (and now waits for it to come up first); without it, semtest/shmtest/msgtest fail just as your notes describe.

aswin-mcw:
> For the record: https://github.com/msys2/msys2-runtime/compare/dscho:also-run-cygwin-testsuite-in-ci demonstrates the commit structure that I would strongly prefer.

Thanks for the reference branch — I've restructured to match it: the fixes are fixup! commits targeting the right topic commits, line endings are enforced via .gitattributes rather than ad-hoc, and the testsuite runs as a job in build.yaml. The only additions on top are the two XFAIL commits (with rationale) and small CI-plumbing fixes for cygserver startup and the worktree artifact.

